PR #7, opened 2016-01-20
Fix installation requirements

### LINKED ISSUE #6: New install doesn't pull requirements
```sh
$ pip install tasktiger
Collecting tasktiger
Installing collected packages: tasktiger
Successfully installed tasktiger-0.2

$ tasktiger
Traceback (most recent call last):
  File "/Users/nolan/.virtualenvs/tt2/bin/tasktiger", line 7, in <module>
    from tasktiger import run_worker
  File "/Users/nolan/.virtualenvs/tt2/lib/python2.7/site-packages/tasktiger/__init__.py", line 2, in <module>
    import click
ImportError: No module named click
```
The same occurs after cloning the repository and running `python setup.py install`. It appears `setup.py` is missing an `install_requires` directive for those libraries in `requirements.txt`.

NPrescott:
Both installing from pip and cloning the repository and running
`setup.py install` both fail as a result of missing requirements. This
patch adds those requirements listed in requirements.txt to setup.py.

Tested with `python setup.py install` as well as `pip install -e ~/tasktiger`

Closes #6

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,11 @@ setup(
     test_suite='tests',
     tests_require=['redis'],
     platforms='any',
+    install_requires=[
+        'click==4.0',
+        'redis==2.10.3',
+        'structlog==15.1.0'
+    ],
     classifiers=[
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
--- a/setup.py
+++ b/setup.py
@@ -10,9 +10,9 @@ setup(
     tests_require=['redis'],
     platforms='any',
     install_requires=[
-        'click==4.0',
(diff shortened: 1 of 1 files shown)


thomasst:
Thanks for the PR! I'm not sure if we should specify exact version numbers in `setup.py` or rather leave them open, see e.g. https://caremad.io/2013/07/setup-vs-requirement/

NPrescott:
Thanks for the quick response, I've updated the PR to remove explicit versioning. Let me know if you'd prefer these two commits be squashed.

thomasst:
Looks good. Thanks!

PR #8, opened 2016-01-21
WiP: Always Eager for when=...

lucasvo:


thomasst:
Would it make sense to compare `dt.utctimetuple()` so we don't need `pytz` as a dependency?

tasktiger/__init__.py
```diff
@@ -276,7 +277,8 @@ def delay(self, func, args=None, kwargs=None, queue=None,
         Queues a task. See README.rst for an explanation of the options.
         """
 
-        if self.config['ALWAYS_EAGER']:
+        if self.config['ALWAYS_EAGER'] and not \
+                (when and when.replace(tzinfo=pytz.UTC) >= datetime.datetime.now(tz=pytz.UTC)):
```
lucasvo: This was just to get it to work. I will make the comparison a bit smarter (not to replace the timezone if there already is one set)

lucasvo:
Yes. I don't want the pytz dependency in there.

PR #9, opened 2016-02-01
Fix task state transitions for unique tasks

thomasst:
This fixes a race condition where a unique task could be queued, moved by the first worker to the active queue, queued again, moved by the second worker to the active queue, and then removed by the first worker before it finished processing on the second worker, raising a "not found" error on that second worker.

The fix for this problem is not trivial and is implemented by 1) disallowing multiple unique tasks being processed at the same time (a unique task is requeued if another instance is being processed), 2) making sure all task transitions behave accordingly if the task is in the destination state.

As part of this, improvements were made to Redis scripts which includes adding unit tests and more documentation.

### DIFF
--- a/tasktiger/redis_scripts.py
+++ b/tasktiger/redis_scripts.py
@@ -1,6 +1,5 @@
 # ARGV = { score, member }
 ZADD_NOUPDATE_TEMPLATE = """
-    local zadd_result
     if not redis.call('zscore', {key}, {member}) then
         {ret} redis.call('zadd', {key}, {score}, {member})
     end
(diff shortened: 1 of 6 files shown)


tasktiger/redis_scripts.py
```diff
@@ -1,36 +1,175 @@
-# ARGV = { score member }
-ZADD_NOUPDATE = """
-    if not redis.call('zscore', KEYS[1], ARGV[2]) then
-        return redis.call('zadd', KEYS[1], ARGV[1], ARGV[2])
+# ARGV = { score, member }
+ZADD_NOUPDATE_TEMPLATE = """
+    local zadd_result
```
wojcikstefan: unnecessary?

thomasst: Yep, thanks, will remove.

PR #10, opened 2016-02-01
include task's args and kwargs in the structlog

wojcikstefan:
This is essential to debugging failed tasks.


wojcikstefan:
nvm, you can just peek the task based on the logged task_id

PR #12, opened 2016-02-17
Task refactor

### LINKED ISSUE #11: ALWAYS_EAGER should serialize/deserialize arguments and everything put into the task object
The behavior of ALWAYS_EAGER is to just call the method directly without json serialize/deserialize. This can lead to bugs that are due to the json serialization not being caught with unit tests.

thomasst:
Refactored the code to make use of the Task class, which now can be used to queue new tasks as well as cancel/retry/inspect them, and is used internally as well.

Added a TaskNotFound exception which is raised when a task cannot be found or does not exist in the expected queue, so we know when e.g. canceling a scheduled task failed.

This also fixes #11.

@closeio/engineering review?

### DIFF
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -165,10 +165,10 @@ class Task(object):
 
     def _move(self, from_state=None, to_state=None, when=None, mode=None):
         """
-        Internal helper to move a task from one state another (e.g. from QUEUED
-        to DELAYED). The "when" argument indicates the timestamp of the task in
-        the new state. If no to_state is specified, the task will be simply
-        removed from the original state.
+        Internal helper to move a task from one state to another (e.g. from
+        QUEUED to DELAYED). The "when" argument indicates the timestamp of the
+        task in the new state. If no to_state is specified, the task will be
+        simply removed from the original state.
 
         The "mode" param can be specified to define how the timestamp in the
         new state should be updated and is passed to the ZADD Redis script (see
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -1,30 +1,287 @@
+import calendar
 import datetime
 import json
-from ._internal import ERROR, QUEUED
+import redis
+import time
 
-# This module contains helper methods to inspect and requeue tasks from queues.
-# TODO: Use task objects wherever we can.
+from ._internal import *
+from .exceptions import TaskNotFound
+
+__all__ = ['Task']
 
 class Task(object):
-    def __init__(self, tiger, queue, state, _data, _ts=None, _executions=None):
+    def __init__(self, tiger, func=None, args=None, kwargs=None, queue=None,
+                 hard_timeout=None, unique=None, lock=None, lock_key=None,
+                 retry=None, retry_on=None, retry_method=None,
+                 _data=None, _state=None, _ts=None, _executions=None):
         """
-        Initializes a Task from the internal structure. Use tasks_from_queue to
-        retrieve tasks from a given queue. Don't call this directly.
+        Queues a task. See README.rst for an explanation of the options.
         """
+
+        if func and queue is None:
+            queue = getattr(func, '_task_queue', tiger.config['DEFAULT_QUEUE'])
+
         self.tiger = tiger
-        self.queue = queue
-        self.state = state
-        self.data = json.loads(_data)
-        if _executions:
-            self.executions = [json.loads(e) for e in _executions if e]
+        self._func = func
+        self._queue = queue
+        self._state = _state
+        self._ts = _ts
+        self._executions = _executions or []
+
+        # Internal initialization based on raw data.
+        if _data is not None:
+            self._data = _data
+            return
+
+        assert func
+
+        serialized_name = serialize_func_name(func)
+
+        if unique is None:
+            unique = getattr(func, '_task_unique', False)
+
+        if lock is None:
+            lock = getattr(func, '_task_lock', False)
+
+        if lock_key is None:
+            lock_key = getattr(func, '_task_lock_key', None)
+
+        if retry is None:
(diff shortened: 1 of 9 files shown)


tests/__init__.py
```diff
@@ -551,5 +552,106 @@ def test_batch_lock_key(self):
...
+        self._ensure_queues()
+
+        # Ensure task is immediately executed.
+        task = Task(self.tiger, exception_task)
+        self.assertRaises(StandardError, task.delay)
+        self._ensure_queues()
+
+        # Ensure there is an exception if we can't serialize the task.
+        task = Task(self.tiger, decorated_task,
+                    args=[datetime.datetime.utcnow()])
```
wojcikstefan: Could add a comment why the task is not serializable in this example (I assume it's bcos datetime is not json-serializable by default, but it's just a hunch that's not clear from just looking at this code).

tests/__init__.py
```diff
@@ -551,5 +552,106 @@ def test_batch_lock_key(self):
...
+
+        # Task.from_id raises if it doesn't exist.
+        self.assertRaises(TaskNotFound, fetch_task)
+
+    def test_delay_scheduled_3(self):
+        task = Task(self.tiger, simple_task, unique=True)
+        task.delay(when=datetime.timedelta(minutes=5))
+        self._ensure_queues(scheduled={'default': 1})
+
+        # We can look up a non-unique task by recreating it.
```
wojcikstefan: s/non-unique/unique

tests/__init__.py
```diff
@@ -551,5 +552,106 @@ def test_batch_lock_key(self):
...
+    def test_delay_scheduled_2(self):
+        task = Task(self.tiger, simple_task, queue='a')
+        task.delay(when=datetime.timedelta(minutes=5))
+        self._ensure_queues(scheduled={'a': 1})
+
+        task_id = task.id
+
+        # We can't look up a non-unique task by recreating it.
+        task = Task(self.tiger, simple_task, queue='a')
+        self.assertRaises(TaskNotFound, task.cancel)
```
wojcikstefan: The exception name is a bit misleading in this case - `TaskNotFound` doesn't actually mean that the task wasn't found here - it means that it can't be identified w/o providing its ID. Maybe it would be better to raise `TaskLookupError` or something like that in this case?

wojcikstefan: Aha, now I see that *every* task has an ID that's generated when you initialize it (and unique tasks always generate the same ID for the same func/args/kwargs). In such case, TaskNotFound makes sense because we try to look the task up via a newly generated non-unique ID.

However, I wonder if we could do a performance optimization where, if the task wasn't initialized via `from_id`, we immediately raise a `TaskNotFound` or a `TaskLookupError` w/o even communicating with Redis.

thomasst: > However, I wonder if we could do a performance optimization where, if the task wasn't initialized via from_id, we immediately raise a TaskNotFound or a TaskLookupError w/o even communicating with Redis.

This is not necessarily the case. Consider this:

```
task = Task(self.tiger, simple_task, queue='a')
taks.delay(when=future)
task.cancel()
```

wojcikstefan: The difference between this scenario and the one in the test is that this scenario actually commits a given ID to Redis. The previous scenario generates a random ID during initialization and then tries to retrieve it.

I guess it's not worth a hassle/code complexity to differentiate between all these scenarios just to avoid a quick Redis call (though I still think it could result in a more obvious exception).

thomasst: At this point it's probably not worth it changing the behavior, but we can potentially revisit this later.

tasktiger/task.py
```diff
@@ -1,30 +1,289 @@
...
+        """
+        return self._ts
+
+    @property
+    def executions(self):
+        return self._executions
+
+    def _move(self, from_state=None, to_state=None, when=None, mode=None):
+        """
+        Internal helper to move a task from one state another (e.g. from QUEUED
```
wojcikstefan: one state *to* another

tasktiger/task.py
```diff
@@ -1,30 +1,289 @@
...
+        now = time.time()
+        self._data['time_last_queued'] = now
+        serialized_task = json.dumps(self._data)
+
+        # convert timedelta to datetime
+        if isinstance(when, datetime.timedelta):
+            when = datetime.datetime.utcnow() + when
+
+        # convert to unixtime
+        if isinstance(when, datetime.datetime):
```
wojcikstefan: Is there a scenario in which `when` is not `datetime.datetime` by now AND is a valid input? If not, maybe we should assert that `isinstance(when, (datetime.datetime, datetime.timedelta))` in the beginning of this method?

thomasst: Yeah we can probably make that assertion. I saw there's actually a bug with `ALWAYS_EAGER` below right now when you specify `when`.

tasktiger/task.py
```diff
@@ -1,30 +1,289 @@
...
+            when = calendar.timegm(when.utctimetuple()) + when.microsecond/1.e6
```
wojcikstefan: could use a comment why `+ when.microsecond/1.e6`

tasktiger/task.py
thomasst: @wojcikstefan see https://github.com/closeio/tasktiger/commit/a55e972d8550174d57d66704c65e8e23d2262941

I decided against having an explicit isinstance check since this is not very pythonic.

lucasvo:
Is this fully backwards compatible? should we start a CHANGELOG.md?

thomasst:
> Is this fully backwards compatible?

Yes, no changes should be required in client code.

PR #14, opened 2016-03-21
Add python3 support

charliewolf:


tasktiger/_internal.py
```diff
@@ -28,7 +30,7 @@ def gen_id():
     """
     Generates and returns a random hex-encoded 256-bit unique ID.
     """
-    return os.urandom(32).encode('hex')
+    return binascii.b2a_hex(os.urandom(32)).decode('utf8')
```
wojcikstefan: What's the advantage here?

tasktiger/task.py
```diff
@@ -228,7 +229,7 @@ def _move(self, from_state=None, to_state=None, when=None, mode=None):
         try:
             pipeline.execute()
         except redis.ResponseError as e:
-            if '<FAIL_IF_NOT_IN_ZSET>' in unicode(e):
+            if '<FAIL_IF_NOT_IN_ZSET>' in text_type(e):
```
wojcikstefan: Does this work if the exception message contains unicode chars?

charliewolf: @wojcikstefan Yes, as text_type == unicode on python2 and str (which is the same as unicode) in python3

tasktiger/_internal.py
charliewolf: The advantage is that it works... encode('hex') is not available in python3

thomasst:
Thanks! Will review once I find the time.

philfreo:
We should add `tox` support for easy testing on multiple versions

charliewolf:
@philfreo I added it!

charliewolf:
Seems to have some errors on py3.4 (circle ci's latest) compared to py3.5 (what I'm running). Need to investigate these.

charliewolf:
@thomasst I fixed the py3.4 issues.

tasktiger/_internal.py
```diff
@@ -1,9 +1,11 @@
+import binascii
 import calendar
 import datetime
 import importlib
 import hashlib
 import json
 import os
+import six
```
thomasst: Why are we importing `six` in this file?

tasktiger/worker.py
```diff
@@ -131,7 +132,7 @@ def _update_queue_set(self, timeout=None):
             r, w, x = select.select([fileno], [], [],
                                     0 if self._queue_set else timeout)
             if fileno in r: # or not self._queue_set:
-                message = gen.next()
+                message = six.next(gen)
```
thomasst: We can just use `next(gen)`, right? I don't care about support for Python older than 2.7.

tasktiger/task.py
thomasst: `if '<FAIL_IF_NOT_IN_ZSET>' in e.args[0]:` should work, too, and wouldn't require `six`.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 click==4.0
 redis==2.10.3
 structlog==15.1.0
+six==1.10.0
```
thomasst: I propose we don't use `six`. See my other comments for reasons.

charliewolf: @thomasst done.

thomasst:
Thanks, this looks good. Will do some more testing and then merge it if everything looks good.

PR #17, opened 2016-05-28
Ability to access the current task object from within the task function

thomasst:
Ability to access the current task object from within the task function using `tiger.current_task` (or `tiger.current_tasks` in case of a batch function).

### DIFF
--- a/tasktiger/_internal.py
+++ b/tasktiger/_internal.py
@@ -15,6 +15,14 @@ ACTIVE = 'active'
 SCHEDULED = 'scheduled'
 ERROR = 'error'
 
+# Global task context. We store this globally (and not on the TaskTiger
+# instance) for consistent results just in case the user has multiple TaskTiger
+# instances.
+g = {
+    'current_task_is_batch': None,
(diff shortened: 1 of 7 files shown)


wojcikstefan:
LGTM, though I think the tests' structure is pretty confusing - it's not immediately clear that `current_task` and `current_tasks` come from `tests/tasks.py` (don't see any explicit imports).

tasktiger/_internal.py
```diff
@@ -15,6 +15,14 @@
 SCHEDULED = 'scheduled'
 ERROR = 'error'
 
+# Global task context. We store this globally (and not on the TaskTiger
+# instance) for consistent results just in case the user has multiple TaskTiger
```
wojcikstefan: In case of multiple intances, is there a way to know which task in `current_tasks` belongs to which TaskTiger instance?

thomasst: It's irrelevant, since `current_task` within a task would always refer to the currently processed task.

anemitz:
@thomasst bump

PR #19, opened 2016-06-05
Update worker.py

rrajaravi:
dictionary comprehension doesn't work in 2.6 version and added compatibility by replacing { } with dict(), hence comprehension would work in 2.6 version


thomasst:
Thanks for the PR! However, I don't think it's reasonable to support Python 2.6 anymore. The Python 2.6 release is officially retired. I highly recommend upgrading to at least Python 2.7.

Issue #21, opened 2016-07-14
Option to use pickle instead of json for [de]serialization


charliewolf:
This would work better for some types, i.e. datetime, etc.

thomasst:
Instead of pickle, how about being able to specify a custom `JSONEncoder`/`JSONDecoder`?

charliewolf:
That would probably be good to have too. That said, `pickle` would be easier out-of-the-box.

thomasst:
I'm not a fan of implementing pickle since it allows code execution (`Warning The pickle module is not secure against erroneous or maliciously constructed data. Never unpickle data received from an untrusted or unauthenticated source.`), but I'm fine with custom JSON encoders/decoders. Patch welcome :)

tkram01:
We would like the option to use pickle as well. Some swig wrappers support pickle serialization and not JSON.

wojcikstefan:
Personally, I don't think there's a point in implementing an insecure serialization/de-serialization option, especially given how often tasks are called with params containing user-generated data. 

@tkram01 you could probably write a very simple middleware that unpickles an object you get from SWIG and then serializes it into JSON before passing it to a TaskTiger task.

philfreo:
(See also #22)

thomasst:
Yeah, I'll review this later, but we can probably do something like #22.

charliewolf:
@wojcikstefan @thomasst I'm not seeing how Pickle is a security issue if the python app is control of pickling/unpickling. My understanding was that the only secufity issue is if you were to accept arbitrary pickled data from users but Tasktiger doesn't do that since it does the serialization/deserialization itself.

thomasst:
That's correct, but it's bad design and unnecessarily opens up a potential point of attack. It's like saying you can use `gets()` in C because the input size is known.

That being said, I don't mind having an option for a customer (de)serializer, as long as we don't ship pickle with TaskTiger (but the user is free to implement/configure it).

alanhamlett:
> That being said, I don't mind having an option for a customer (de)serializer, as long as we don't ship pickle with TaskTiger (but the user is free to implement/configure it).

And just point people in the right direction with a custom JSONEncoder/Decoder example.

PR #22, opened 2016-07-19
Custom serializers

charliewolf:


charliewolf:
Addresses #21

tasktiger/task.py
```diff
@@ -12,7 +11,7 @@ class Task(object):
     def __init__(self, tiger, func=None, args=None, kwargs=None, queue=None,
                  hard_timeout=None, unique=None, lock=None, lock_key=None,
                  retry=None, retry_on=None, retry_method=None,
-                 _data=None, _state=None, _ts=None, _executions=None):
+                 _data=None, _state=None, _ts=None, _executions=None, deserialize=json.loads, serialize=json.dumps):
```
thomasst: I think we can remove these.

tasktiger/__init__.py
```diff
@@ -132,10 +132,17 @@ def __init__(self, connection=None, config=None, setup_structlog=False):
 
             # If non-empty, a worker only processeses the given queues.
             'ONLY_QUEUES': [],
+
```
thomasst: Would be nice to have a quick comment just like the other settings.

tests/__init__.py
```diff
@@ -1,3 +1,4 @@
+import pickle
```
thomasst: I preferably don't want to mention `pickle` in this package since it suggests that this would be a sane choice. A better example would be JSON with a custom `JSONEncoder`/`JSONDecoder`. This would then also ensure that the given serialization is actually being used (since the default encoder would throw an error).

PR #23, opened 2016-07-19
Fix exceptions in python3

charliewolf:
It seems python3 doesn't like it when you try to get the traceback by doing `sys.exc_info()` outside the try/except block. Moving it there appears to resolve the issue.


charliewolf:
Closes #25

charliewolf:
@thomasst ping

PR #27, opened 2016-09-26
Properly check for absent execution before use

tsx:


thomasst:
Note that the log statement below will also fail if `execution` is `None`. Is there a specific flow that can cause `execution` to be `None`, so we can test for it?

thomasst:
@tsx I added a test for this, in case you haven't seen.

PR #28, opened 2016-11-08
Periodic tasks

thomasst:
No more reasons to use celery.

@closeio/engineering low priority, but feel free to review.


README.rst
```diff
@@ -335,6 +336,30 @@ The following options can be only specified in the task decorator:
...
+    argument, followed by the additional args.
+
+  The schedule function must return the next task execution datetime, or
+  ``None`` to prevent periodic execution. The function is executed to determine
+  the initial taks execution date when a worker is initialized, and to determine
+  the next execution date when the task is about to get executed.
+
+  For most common scenarios, the ``periodic`` built-in function can be passed:
+
+  - ``periodic(seconds=0, minutes=0, hours=0, days=0, weeks=0, start_date=None,
```
tsx: For larger intervals (days, weeks) it might be useful to have a flag like `run_on_weekends=True` and specify time of the day to start execution. For example, someone wants to run some cleanup on only on weekdays at 3am.

thomasst: I can see that this is a possibility. But for now, I want to keep this simple, and if someone needs different schedules, they can write a custom scheduler function. If needs arise, we can add more complex schedulers at a later point in time.

PR #33, opened 2016-11-14
Handle ECHILD error in os.waitpid()

thomasst:
See the comment of a description on how it happens. We will most likely see a "encountered ECHILD while os.waitpid" error once a week or so, which will give us more insight on whether `return_code` is ever `None`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,8 @@
+machine:
+  services:
+    - redis
+
 test:
   override:
+    - pyenv global 2.7.12 3.4.4 3.5.2
     - tox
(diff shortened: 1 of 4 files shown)


thomasst:
@tsx pls review

tsx:
A few lines below:

`status = not return_code`

With return_code == None, this will assume successful execution. Isn't it more safe to assume failed execution and retry the task? (not entirely sure, just asking)

Otherwise LGTM.

thomasst:
Good point! Will fix

PR #35, opened 2016-11-30
Make function serialization and importing less naive

### LINKED ISSUE #26: classmethods aren't called
If you pass `module.Foo.bar` (where `Foo.bar` is a `@classmethod` of `Foo`) it tries to call `module.bar`

charliewolf:
This resolves #26


thomasst:
Thanks! This looks good. Now we just need a unit test that shows the new functionality.

Fun discovery: If you have `f = lambda: 123` in a module, both `__name__` and `__qualname__` return `"<lambda>"` instead of `"f"`. But couldn't find a way to make it work with lambdas.

Are there any conventions to separate module name and function name, or did you chose `|` arbitrarily? E.g. pytest uses `::` as a separator, and nosetests uses `:` to run a specific test. I'm leaning towards using `:`. Do we need the special case when storing `exception_name` for backwards compatibility, or should we just use the new separator?

charliewolf:
@thomasst See the updates I made per your comments! Thanks and happy holidays.

thomasst:
Added https://github.com/closeio/tasktiger/commit/4f77de7be89d988264c50fea09a3747d67c4eb0c for forward-compatibility in 0.8.4. Will merge this PR for 0.9.

charliewolf:
@thomasst I went ahead and rebased this to fix the merge conflicts that popped up from your test refactoring.

thomasst:
Thanks! Going to release 0.9 soon.

PR #37, opened 2016-12-08
Fix handling of "not found" error if task disappears

thomasst:
Ensure that a task object that disappears while the task is processing is handled properly. This could happen when a worker processes a task, then hangs for a long time, causing another worker to pick up and finish the task. Then, when the original worker resumes, the task object will be gone. Make sure we log a "not found" error and move on.


wojcikstefan:
LGTM!

PR #47, opened 2017-01-02
README: fix usage of custom redis connection in python 3

macobo:
redis.py [docs](https://github.com/andymccurdy/redis-py) state: 

> If the client's decode_responses flag is set the False (the default), the 'channel', 'pattern' and 'data' values in message dictionaries will be byte strings (str on Python 2, bytes on Python 3).

tasktiger seems to assume for what I think is compatibility reasons, that redis-py returns plain strings. This PR makes that explicit in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -137,7 +137,8 @@ decorate and queue tasks. The constructor takes the following arguments:
 
 - ``connection``
 
-  Redis connection object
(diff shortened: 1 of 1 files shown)


thomasst:
Thanks!

PR #48, opened 2017-01-03
Add CLI commands

jkemp101:
Initial work on getting queue stats and dumping tasks to clear queues.

### DIFF
--- a/tasktiger/cli.py
+++ b/tasktiger/cli.py
@@ -1,0 +1,133 @@
+"""TaskTiger CLI."""
+
+import argparse
+import datetime
+import json
+import sqlite3
+
+from _internal import QUEUED, ACTIVE, SCHEDULED, ERROR
+from task import Task
+
+
+class TaskTigerCLI(object):
+    """TaskTiger CLI class."""
+
+    def __init__(self, tiger, args):
+        self.tiger = tiger
+        self._parse_args(args)
+
+    def _parse_args(self, args):
+        parser = argparse.ArgumentParser()
+        subparsers = parser.add_subparsers(dest='command')
+
+        common_parser = argparse.ArgumentParser(add_help=False)
+
+        # Queue stats command
+        subparsers.add_parser('queue_stats', help='Print queue statistics.',
+                              parents=[common_parser])
+
+        # Dump queue command
+        dq_parser = subparsers.add_parser('dump_queue', help='Dump tasks to sqlite database and clear queue.',
+                                          parents=[common_parser])
+        dq_parser.add_argument('-q', '--queue', help='Queue name', required=True)
+        dq_parser.add_argument('-f', '--file', help='Sqlite3 file name', required=True)
+        dq_parser.add_argument('-s', '--state', help='Task state (default=queued)',
+                               required=False,  choices=[QUEUED, SCHEDULED], default='queued')
+        dq_parser.add_argument('-b', '--batches', help='Number of batches to dump, defaults to all tasks',
+                               required=False, default=-1)
+
+        # Sample queue command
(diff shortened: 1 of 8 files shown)


jkemp101:
@closeio/engineering This is ready for review.  The loading of tasks back into the queues is just stubbed out right now, that can be finished later.  

Examples of how to dump tasks: https://github.com/closeio/tasktiger/pull/48/files#diff-75f4fd64deca9f40437ffab6fdf4657dR40

tasktiger/cli.py
```diff
@@ -0,0 +1,235 @@
+"""TaskTiger CLI."""
+
+import argparse
```
thomasst: We're already using `click` for the `run_worker` CLI, so we should use that instead of `argparse`.

tests/test_cli.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Dump queue to sqlite3 file
+        cli = TaskTigerCLI(self.tiger, ['dump_queue', '-f', db_file, '-q', 'cli', '-s', 'queued'])
+        cli.run()
+
+        # Verify contents of tasks table
+        conn = sqlite3.connect(db_file)
+        cursor = conn.cursor().execute('select id, data from tasks where id=?', (task.id, ))
+        row = cursor.fetchone()
+        self.assertIsNotNone(row)
```
thomasst: Should use standard Python `assert` since we're using pytest (here and in other instances).

tasktiger/task.py
```diff
@@ -418,11 +418,12 @@ def cancel(self):
         """
         self._move(from_state=SCHEDULED)
 
-    def delete(self):
+    def delete(self, from_state=ERROR):
```
thomasst: The "public" `Task` API doesn't let you specify this on purpose: If it's a scheduled task, you can `cancel()` it, and if it's an errored out task you can `delete()` it, but it's unsafe to delete active tasks. It should be okay to delete queued tasks though. The tricky part is that technically a task could be in multiple states at the same time (there's nothing that prevents a unique task to be in multiple queues), and depending on how it was retrieved it may not know the state here. That's one reason on why we have two methods to get rid of a task.

Maybe we want to change the API to be simply `delete(self, from_state=None)` working as follows: `from_state = from_state or self.state; assert from_state in (QUEUED, SCHEDULED, ERROR)`, and deprecate the `cancel()` method.

jkemp101:
I think the CLI should be part of this module.  I was actually thinking once we get this working as a flask script command we should also create a `tasktigercli` in `bin` just like redis gives you a `redis-cli` and celery gives you the `celery` command.

tasktiger/task.py
jkemp101: I took the first crack at this.  See what you think.

thomasst:
> I think the CLI should be part of this module.

I think it's a bit different from Redis' or Celery's CLI because it's not really an interface to TaskTiger, but  (at least at this stage) mostly just an importer/exporter to SQL, which seems like a separate component to me.

jkemp101:
I think it already has some beneficial use cases and we should think about what we want this to become not necessarily what it is today.

__List Queues__  
1. `./manage.py tasktigercli queue_stats`

__Purge a Queue__  
1. `./manage.py tasktigercli dump_queue -q testb -s queued -p`

__Move tasks from `testa` queue to the `testb` queue__  
1. `./manage.py tasktigercli dump_queue -q testa -s queued -f /tmp/testa.db`
2. `./manage.py tasktigercli load_queue -q testb -s queued -f /tmp/testa.db` (This command isn't complete yet)

thomasst:
We'll need this at some point -- let's keep it open until we find time to work on this.

PR #49, opened 2017-01-04
Refactor to use get_message for queue_set updates

jkemp101:
The existing logic would not process all pending messages.  This change should ensure between each run we have an up to date set of queues.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 click==4.0
-redis==2.10.3
+redis==2.10.5
 structlog==15.1.0
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -138,21 +138,12 @@ class Worker(object):
         the activity channel were read.
         """
 
-        # Pubsub messages generator
-        gen = self._pubsub.listen()
-        while True:
-            # Since Redis' listen method blocks, we use select to inspect the
-            # underlying socket to see if there is activity.
-            fileno = self._pubsub.connection._sock.fileno()
-            r, w, x = select.select([fileno], [], [],
-                                    0 if self._queue_set else timeout)
-            if fileno in r: # or not self._queue_set:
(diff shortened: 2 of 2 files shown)


thomasst:
The code looks cleaner, but how is this different from the original logic?

jkemp101:
The original code seems to depend on `select` to determine if there are additional messages available.  But `select` will only tell you if there is data available in the OS TCP buffer.  So I think this is what happens:

1. Queues are basically empty
1. 100 slow running assign_email tasks go in the queue
1. 1 regular webhook task is added to the queue
1. The current code uses `select` to see there is data in the TCP receive buffer and makes the first `next(gen)` call.
1. `next(gen)` moves all data from the TCP receive buffer into a python data structure and returns 1 message
1. Code looks at the 1 message and adds the `assign_email` queue to `_queue_set`.
1. The next `select` call indicates there is no data in the TCP receive buffer so the loop exits.  This second call to `select` does not tell you there are no messages pending.  This is the issue, there are many messages pending.
1. The next 99 assign_email messages are not processed so we don't learn about the webhook queue for a while.

So now:
 * Each "batch" of new messages result in a single message being processed which may uncover a new queue
 * This tasktiger process will now pretty much be dedicated to the assign_email queue and doesn't know that a webhook is pending.  The `random.shuffle(queues)` is shuffling a single queue.

Its easy to reproduce.  Just put a `print queues` after the shuffle.  Start 1 tasktiger process and pump in 100 slow tasks on 1 queue and 100 on another queue.  You will see its just keeps shuffling the first queue.

thomasst:
Makes sense. The key here is that `get_message()` calls `can_read()` under the hood, which does `return self._parser.can_read() or bool(select([sock], [], [], timeout)[0])`, i.e. it checks the read buffer as well. Great job for finding this!

PR #52, opened 2017-01-09
Only requeue expired tasks that are retriable on JobTimeoutException

thomasst:
This prevents retrying tasks that are not idempotent in case the worker crashes. We now also do the requeueing periodically using a lock and are more efficient about it (no per-queue round trip to Redis since finding expired tasks is now handled in a Lua script).

- [x] Unit tests that illustrate the new behavior.

@closeio/engineering feel free to review

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -15,7 +15,7 @@ import traceback
 from .redis_lock import Lock
 
 from ._internal import *
-from .exceptions import RetryException
+from .exceptions import RetryException, TaskNotFound
 from .retry import *
 from .stats import StatsThread
 from .task import Task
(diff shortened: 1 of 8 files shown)


jkemp101:
LGTM

tasktiger/worker.py
```diff
@@ -149,30 +149,66 @@ def _worker_queue_expired_tasks(self):
...
+
+        # Note that we use the lock both to unnecessarily prevent multiple
+        # workers from requeueing expired tasks, as well as to space out
+        # this task (i.e. we don't release the lock unless we've exhausted our
+        # batch size, which will hopefully never happen)
+        lock = Lock(self.connection, self._key('lock', 'queue_expired_tasks'),
+                    timeout=self.config['REQUEUE_EXPIRED_TASKS_INTERVAL'])
+
+        acquired = lock.acquire(blocking=False)
+        if acquired:
```
wojcikstefan: This would be nicer in my opinion:

```py
if not acquired:
    return
```

tasktiger/worker.py
```diff
@@ -149,30 +149,66 @@ def _worker_queue_expired_tasks(self):
...
+
+                    # XXX: Ideally, the following block should be atomic.
+                    if not self.connection.get(self._key('task', task_id)):
+                        self.log.error('not found',
+                                       queue=queue, task_id=task_id)
+                        task = Task(self.tiger, queue=queue,
+                                    _data={'id': task_id}, _state=ACTIVE)
+                        task._move()
+
+            # Keep requeueing if we've exhausted our batch size, otherwise
```
wojcikstefan: "Keep requeueing" is a bit confusing in the context of this method. Why is releasing the lock equivalent with requeueing?

tests/tasks.py
```diff
@@ -3,7 +3,7 @@
 
 import redis
 
-from tasktiger import RetryException
+from tasktiger import JobTimeoutException, RetryException
```
wojcikstefan: Unnecessary import?

tests/test_base.py
```diff
@@ -787,6 +787,46 @@ def test_current_tasks_eager(self):
 class ReliabilityTestCase(BaseTestCase):
     """ Test behavior if things go wrong. """
```
wojcikstefan: Per https://www.python.org/dev/peps/pep-0257/#one-line-docstrings:

> There's no blank line either before or after the docstring.

tests/test_base.py
```diff
@@ -787,6 +787,46 @@ def test_current_tasks_eager(self):
...
 
+    def _test_expired_task(self, task, expected_state):
+        """
+        Ensure the given task ends up in the expected state if the worker is
+        killed prematurely.
```
wojcikstefan: Could add a note that `task` has to run for longer than `DELAY` for this test to work.

wojcikstefan:
LGTM

tasktiger/worker.py
thomasst: Normally, if we don't release the lock, it expires after its timeout. As stated in the comment above, "we use the lock both to unnecessarily prevent multiple workers from requeueing expired tasks, as well as to space out this task (i.e. we don't release the lock unless we've exhausted our batch size, which will hopefully never happen)". So if there are more expired tasks left to requeue, the next worker executing `_worker_queue_expired_tasks` will be able to immediately pick up instead of encountering a blocked lock.

wojcikstefan: Thanks, that makes sense. Perhaps a clearer way to say it is something along these lines:
```
# Release the lock immediately if we processed a full batch. This way, another
# process will be able to pick up another batch immediately without waiting
# for the lock to time out.
```

PR #66, opened 2017-05-24
Add task check interval

jkemp101:


jkemp101:
Fixing under another PR

PR #70, opened 2017-05-26
Improve redis usage

jkemp101:
Reduces Redis activity by:
1. Support lockable queues to prevent multiple workers from processing the same queue
1. Sleep when possible between worker loop execution

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -151,6 +151,8 @@ class TaskTiger(object):
 
             # If non-empty, a worker excludes the given queues from processing.
             'EXCLUDE_QUEUES': [],
+
+            'SINGLE_WORKER_QUEUES': [],
         }
         if config:
             self.config.update(config)
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -105,7 +105,7 @@ class TaskTiger(object):
             # task, use the following default method.
             'DEFAULT_RETRY_METHOD': fixed(60, 3),
 
-            # After how many seconds a task that can't require a lock is
+            # After how many seconds a task that can't acquire a lock is
             # retried.
             'LOCK_RETRY': 1,
 
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -142,6 +142,13 @@ class TaskTiger(object):
             # Upper bound for the time it takes to queue all periodic tasks.
             'QUEUE_PERIODIC_TASKS_LOCK_TIMEOUT': 10,
 
+            # Single worker queues can reduce redis activity in some use cases
+            # by locking at the queue level instead of task or task groups.
+            # These queues will only allow a single worker to access the queue
+            # at a time.  Useful in environments with large queues and many
+            # worker processes.
+            'SINGLE_WORKER_QUEUES': [],
+
             # The following settings are only considered if no explicit queues
             # are passed in the command line (or to the queues argument in the
             # run_worker() method).
@@ -151,8 +158,6 @@ class TaskTiger(object):
 
             # If non-empty, a worker excludes the given queues from processing.
             'EXCLUDE_QUEUES': [],
-
-            'SINGLE_WORKER_QUEUES': [],
         }
         if config:
             self.config.update(config)
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -92,7 +92,7 @@ class TaskTiger(object):
             # channel and check for scheduled or expired items.  The batch
             # timeout will delay the specified seconds after the first message
             # to wait for additional messages, useful for very active systems.
-            # Appropriate values: 0 >= SELECT_BATCH_TIMEOUT >= SELECT_TIMEOUT
+            # Appropriate values: 0 >= SELECT_BATCH_TIMEOUT <= SELECT_TIMEOUT
             'SELECT_TIMEOUT': 1,
             'SELECT_BATCH_TIMEOUT': 0,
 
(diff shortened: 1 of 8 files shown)


tasktiger/worker.py
```diff
@@ -514,13 +580,20 @@ def _process_from_queue(self, queue):
...
+                          processed=processed_count)
                 for task in processed_tasks:
                     self._finish_task_processing(queue, task, success)
 
-        return task_ids
+        if queue_lock:
+            queue_lock.release()
+            log.debug('released swq lock')
+
+        return task_ids, processed_count
```
thomasst: Update the docstring of this function to indicate what's being returned.

tasktiger/__init__.py
```diff
@@ -89,8 +89,12 @@ def __init__(self, connection=None, config=None, setup_structlog=False):
             'DEFAULT_QUEUE': 'default',
 
             # After how many seconds time out on listening on the activity
-            # channel and check for scheduled or expired items.
+            # channel and check for scheduled or expired items.  The batch
+            # timeout will delay the specified seconds after the first message
+            # to wait for additional messages, useful for very active systems.
+            # Appropriate values: 0 >= SELECT_BATCH_TIMEOUT <= SELECT_TIMEOUT
```
thomasst: One more time. This should really read:

`0 <= SELECT_BATCH_TIMEOUT <= SELECT_TIMEOUT`

tasktiger/__init__.py
```diff
@@ -89,8 +89,12 @@ def __init__(self, connection=None, config=None, setup_structlog=False):
             'DEFAULT_QUEUE': 'default',
 
             # After how many seconds time out on listening on the activity
-            # channel and check for scheduled or expired items.
+            # channel and check for scheduled or expired items.  The batch
+            # timeout will delay the specified seconds after the first message
+            # to wait for additional messages, useful for very active systems.
+            # Appropriate values: 0 <= SELECT_BATCH_TIMEOUT <= SELECT_TIMEOUT
             'SELECT_TIMEOUT': 1,
+            'SELECT_BATCH_TIMEOUT': 0,
```
thomasst: This timeout applies regardless of whether the queue is a batch queue (in `BATCH_QUEUES`), right?

tasktiger/worker.py
```diff
@@ -428,23 +513,106 @@ def check_child_exit():
...
+        Internal method to process a task batch from the given queue.
+
+        Args:
+            queue: Queue name to be processed
+
+        Returns:
+            Task IDs:   List of tasks that were processed (even if there was an
+                        error so that client code can assume the queue is empty
+                        if nothing was returned)
+            Count:      The number of tasks that were attemped to be executed
```
thomasst: (or -1 if the queue lock couldn't be acquired).

tasktiger/__init__.py
thomasst: Can you elaborate a bit on why this overall helps with high loads? And how is it better than a simple `sleep` in between loop iterations if we haven't processed any tasks in the previous iteration (`not _did_work`)?

tasktiger/worker.py
```diff
@@ -344,11 +429,9 @@ def _execute(self, queue, tasks, log, locks, all_task_ids):
...
             # Main process
             log = log.bind(child_pid=child_pid)
-            log.debug('processing', func=task_func, params=[{
-                    'task_id': task.id,
-                    'args': task.args,
-                    'kwargs': task.kwargs,
-            } for task in tasks])
+            for task in tasks:
+                log.info('processing', func=task_func, task_id=task.id,
+                         params={'args': task.args, 'kwargs': task.kwargs})
```
thomasst: Don't care too much about the log format, but why the change from one log message to multiple?

tests/test_base.py
```diff
@@ -31,12 +31,17 @@ def teardown_method(self, method):
         self.conn.flushdb()
 
     def _ensure_queues(self, queued=None, active=None, error=None,
-                       scheduled=None):
+                       scheduled=None, ignore_queues=[]):
```
thomasst: Maybe we could only register the periodic task in a certain test class so we don't have to include this check everywhere?

tasktiger/__init__.py
jkemp101: Correct, the `BATCH` in that setting has nothing to do with Batch Queues.

tasktiger/worker.py
jkemp101: When relying on json log output you want to be able to filter by `task_id` and see when the task was started, done, and/or errored.  The `processing` event had multiple `task_id`'s in the `params` array which means you need to filter differently for the same task depending on if you are searching for dones and errors versus the processing event.

This isn't really an issue if you are just grepping for a specific task id in a log file but if you are using something like Kibana to do the search it means you can unknowingly miss log entries when you query `task_id=XYZ`.

tasktiger/__init__.py
jkemp101: This deals with the specific use case where you have many workers processing a very active queue.  Assume we start with an empty queue. If you have 20 workers and put 1 task on the queue all 20 workers will detect this as a `new_queue_found` and immediately exit the `_wait_for_new_tasks` loop.  1 worker will successfully move that task from `QUEUED` to `ACTIVE` and 19 workers will fail to move any tasks because `QUEUED` is now empty.  In this very simple case we wasted 19 redis calls.

Now in an active queue where we are constantly queueing tasks, by waiting just a little before exiting the `_wait_for_new_tasks` we let `QUEUED` to fill with a few more tasks so that more that 1 out of the 20 will have a task available to move from `QUEUED` to `ACTIVE`.  Maybe something like `SELECT_FILL_DELAY` would be a better name than `SELECT_BATCH_TIMEOUT`.

I'm not exactly sure what you mean by using a `sleep`.  It sleeps in the `get_message` so that we always get messages off the channel if they are available.  A regular sleep would have the same result of letting `QUEUED` to fill but maybe leaving a few messages on the channel that would be a false positive the next time we entered the loop.

PR #71, opened 2017-06-02
feat: crontab expression for schedule

dawncold:
I use `croniter` to parse crontab expression and provide a `cron_expr` function for `schedule` argument, is it possible add this feature into tasktiger?


thomasst:
Thanks for the PR!

Would it be possible to make this an optional dependency (e.g. inline the imports)?
Also, could you add a test and docs (README / docstring)?

PR #72, opened 2017-06-08
fix: worker process not exit as stats thread is running

dawncold:
Enabled stats (default set `STATS_INTERVAL` to 60), a stats thread is running before connect to `redis`, if connects to `reids` is failed, `worker` main thread will exit, but stats thread is running, this behavior results unexpected `worker` process status.

I use `supervisord` to manage `worker` processes, after I stopped queue redis, all `workers` stops working, but `worker`s process status are running, because stats threads work fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ build/
 dist/
 *.egg-info/
 .tox/
+.env/
+.idea/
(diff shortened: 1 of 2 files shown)


jkemp101:
I set the Stats thread to be a daemon thread in the last commit (https://github.com/closeio/tasktiger/commit/7dd448e483d3ae1edf8e669a5ba76ee1c6ec9e28) which I think should fix this issue.  That thread will no longer keep the process from exiting.

PR #75, opened 2017-06-20
Run tests on Python 3.6.1

alanhamlett:
This change runs tests on Python 3.6 to commit to support 3.6 going forward.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,5 +4,5 @@ machine:
 
 test:
   override:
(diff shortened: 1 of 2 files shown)


thomasst:
Thanks! There's 3.6.2 now in Circle (https://circleci.com/docs/1.0/build-image-trusty/) so I'm merging https://github.com/closeio/tasktiger/pull/86.

PR #79, opened 2017-08-15
Support task.delay(when=...) on decorated functions

tsx:
Support `task.delay(when=datetime.timedelta(seconds=1))`.


thomasst:
I'm closing this since it would break with tasks that take `when` as a kwarg.

PR #81, opened 2017-08-29
fix: executing lua script error since lua script is not loaded

dawncold:
Since latested python redis client [changed lua script loading mechanism](https://github.com/andymccurdy/redis-py/commit/94975d310353a656fb12fd68cdd2f2a4ca5ae502), library pipeline will load scripts before execution, tasktiger needs load these scripts manually


thomasst:
Thanks! Duplicate of https://github.com/closeio/tasktiger/pull/80

PR #85, opened 2017-09-12
Include license and tests in sdist

EdwardBetts:
It would be great to include the test suite in the pypi release tarball.
That way people who download and build the release can run the tests to
ensure the library is working.

Linux distributions, such as Debian, base their packages of Python
modules on the pypi release. This means the test suite can be run when
building a Debian package, it should catch mistakes in the packaging or
errors in dependencies.

I've written some more on this topic on the Debian Python mailing list.
https://lists.debian.org/debian-python/2016/04/msg00074.html

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include tox.ini LICENSE
+recursive-include tests *.py


thomasst:
Thanks! I'm going to merge this. Do you know if `package_data` in setup.py can be replaced by corresponding lines in `MANIFEST.in`, or is this the suggested way?

EdwardBetts:
I think they're too different. `MANIFEST.in` lists files to be include in the .tar.gz generated by `setup.py sdist` while `package_data` specifics files to be installed with `setup.py install`.

PR #88, opened 2017-10-10
use a pipeline to load scripts and run actual pipeline lua script

dawncold:
Modification passed all tests on Python2.7 and Python 3.5

I checked redis AOF file, before executing pipeline lua script, scripts are loaded explicitly, I think it fixes my issue.

`load_scripts()` has two shortcomings:

* It will not load scripts which are existed in Redis script cache, we need script load explicitly, and let Redis writes `SCRIPT LOAD` in AOF file.
* It loads script one by one and calls `SCRIPT LOAD` immediately, we must let all scripts loading and lua script pipeline execution in one transaction.

### DIFF
--- a/tasktiger/redis_scripts.py
+++ b/tasktiger/redis_scripts.py
@@ -450,7 +450,7 @@ class RedisScripts(object):
         # [queue1, task1, queue2, task2] -> [(queue1, task1), (queue2, task2)]
         return list(zip(result[::2], result[1::2]))
 
-    def execute_pipeline(self, pipeline, client=None):
+    def execute_pipeline(self, pipeline):
         """
         Executes the given Redis pipeline as a Lua script. When an error
         occurs, the transaction stops executing, and an exception is raised.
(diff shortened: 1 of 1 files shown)


tasktiger/redis_scripts.py
```diff
@@ -469,20 +469,26 @@ def execute_pipeline(self, pipeline, client=None):
         results = redis_scripts.execute_pipeline(p)
         """
 
+        executing_pipeline = None
         try:
+            executing_pipeline = self.redis.pipeline()
+
+            # Load scripts
+            executing_pipeline.script_load(self._execute_pipeline.script)
```
thomasst: Do we need this? Wouldn't Redis properly keep track of this script since it's the one we're directly calling? And `executing_pipeline.execute()` would ensure it's loaded for the first time since it calls `load_scripts()`.

tasktiger/redis_scripts.py
```diff
@@ -450,7 +450,7 @@ def get_expired_tasks(self, key_prefix, time, batch_size, client=None):
         # [queue1, task1, queue2, task2] -> [(queue1, task1), (queue2, task2)]
         return list(zip(result[::2], result[1::2]))
 
-    def execute_pipeline(self, pipeline, client=None):
+    def execute_pipeline(self, pipeline):
```
thomasst: Let's keep the `client` arg to keep it consistent with the other scripts, and with pyredis' `Script.__call__` function, and to possibly allow third party code to run this with a different Redis client instance.

We could just say `(client or self.redis).pipeline()` below (or `if not client: client = self.redis`).

thomasst:
Thanks for finding this and for the initial fix!

PR #90, opened 2017-10-17
Fix load_executions in task_from_queue with custom redis prefix

ikornaselur:
When loading executions with tasks, a hard coded redis prefix (t:) is used, meaning with a custom redis prefix, the function will crash. This patch will use the `_key` helper function to use the configured custom redis prefix.


ikornaselur:
Fixed with a test in #104

PR #92, opened 2017-10-31
feat: Allow custom serializer and deserializers for task

### LINKED ISSUE #21: Option to use pickle instead of json for [de]serialization

kgritesh:
Fixes #21

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -159,6 +159,13 @@ class TaskTiger(object):
 
             # If non-empty, a worker excludes the given queues from processing.
             'EXCLUDE_QUEUES': [],
+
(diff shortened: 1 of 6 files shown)


xuru:
FYI, this doesn't use the custom encoder/decoder in gen_unique_id which I ran into using this code.

kgritesh:
@alanhamlett @xuru let me know if we need to use this custom serialiser in `get_unique_id` function. Also let me know if anything pending for this to be merged?

xuru:
@kgritesh I took a different approach then passing in a call back for serialization/deserialization.  I decided to do it like flask does it, and pass in the json encoder/decoder in order to serialize custom types.  I created a pull request: #99

PR #99, opened 2017-12-30
Set json encoder/decoder via config

xuru:
Allows for custom serialization/deserialization and defaults to the built json encoder/decoder.  This also sets the encoder/decoder for all places `json.dump(s)` and `json.load(s)` is used (i.e. `gen_unique_id`).

This allows you to use a custom encoder/decoder and still use `unique=True` on tasks.

TODO: add tests, and docs

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -2,6 +2,7 @@ import click
 from collections import defaultdict
 import importlib
 import logging
+from json import JSONDecoder, JSONEncoder
 import redis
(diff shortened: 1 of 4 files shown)


thomasst:
Thanks! Any downside on doing it the way https://github.com/closeio/tasktiger/pull/92 or https://github.com/closeio/tasktiger/pull/22 do it? Those give additional flexibility since they allow for encoders/decoders other than JSON.

Issue #78, opened 2017-07-14
restart queue redis (appendonly yes) would enqueued some processed tasks again


dawncold:
~~I found some tasks which are done successfully are enqueued again.~~

~~I have a transactional sms queue, send sms via another platform like Twilio, I found one sms task is done successfully (found sms record on sms platform), and there is no failed tasks in queue.~~

~~Every night we will backup database and redis aof files, during backup, we will shutdown database and redis, but we don't shutdown tasktiger worker processes, so it will raise exception of redis connection and it will be restarted by `supervisord`, after restart tasktiger several times, some sms tasks are enqueued again even they have been processed successfully.~~

~~I have tested, if I only restart my redis-server, some non-existing tasks will be enqueued as if my redis is rollback.~~

***
Today, we think we have found root cause of this issue.

`Tasktiger` use some `lua` scripts for utilizing `lua` language features, such as `for`, `function`. The customized `pipeline` script calls other `lua` scripts, that will be recorded in `AOF` file like this: 

> execute pipeline script with some sub-scripts: script A SHA1 and args for A, script B SHA1 and args for B, ...

`BGWRITEAOF` will remove `SCRIPT LOAD`, but you can still call these scripts via `EVALSHA` before restart `redis` or `SCRIPT FLUSH`.

When `redis` loading data from `AOF` file, `pipeline` script using `EVALSHA` will be failed because scripts are not loaded, so some tasks are not removed from active queue, `worker` will move them to `queued` and process them again.

We have to `BGREWRITEAOF` every hour and rewrite `AOF` before shutdown `redis`.

thomasst:
Thanks for reporting this! Do we need to call `SCRIPT LOAD` *within* the pipeline Lua script? I haven't tried if this works, but my guess is it would solve the problem.

dawncold:
I think it should be fix if `SCRIPT LOAD` is called before Lua script, and I don't know why Redis don't do this.

thomasst:
We directly call scripts by the SHA sum from the `_G` variable so I don't think Redis would know that it needs to load additional scripts before executing our Lua script.
https://github.com/closeio/tasktiger/blob/master/tasktiger/lua/execute_pipeline.lua#L46

Maybe the solution could be to use a pipeline that first loads all the scripts (regardless of whether they're already loaded), and then runs the actual pipeline Lua script. That way Redis shouldn't break it up.

duxiaoyao:
[...]

PR #100, opened 2018-01-04
Fix pipeline script execution

thomasst:
Supersedes https://github.com/closeio/tasktiger/pull/88 / https://github.com/closeio/tasktiger/pull/98
Fixes https://github.com/closeio/tasktiger/issues/78

For Redis 3.2 and higher, this uses single command replication, which should benefits most pipeline executions and avoids replication issues where scripts referenced in pipelines aren't replicated properly (since now only the resulting commands are replicated).

For older Redis versions, all scripts are loaded in a real Redis pipeline, which causes Redis to always replicate the "SCRIPT LOAD" commands. The downside is that the content of any referenced scripts is sent with every pipeline execution.

Tests require (at least) Redis 3.2.

### DIFF
--- a/tasktiger/redis_scripts.py
+++ b/tasktiger/redis_scripts.py
@@ -297,16 +297,18 @@ class RedisScripts(object):
 
         self._execute_pipeline = self.register_script_from_file('lua/execute_pipeline.lua')
 
-        self.can_replicate_commands = self._can_replicate_commands()
-
-    def _can_replicate_commands(self):
+    @property
+    def can_replicate_commands(self):
         """
         Whether Redis supports single command replication.
         """
-        info = self.redis.info('server')
-        version_info = info['redis_version'].split('.')
-        major, minor = int(version_info[0]), int(version_info[1])
-        return major > 3 or major == 3 and minor >= 2
+        if not hasattr(self, '_can_replicate_commands'):
+            info = self.redis.info('server')
+            version_info = info['redis_version'].split('.')
+            major, minor = int(version_info[0]), int(version_info[1])
+            result = major > 3 or major == 3 and minor >= 2
+            self._can_replicate_commands = result
+        return self._can_replicate_commands
 
     def register_script_from_file(self, filename):
         with open(os.path.join(os.path.dirname(os.path.realpath(__file__)),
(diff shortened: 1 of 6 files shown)


tasktiger/redis_scripts.py
```diff
@@ -469,20 +480,39 @@ def execute_pipeline(self, pipeline, client=None):
...
                 script_args += [len(args)-1] + list(args)
 
-            # Make sure scripts exist
-            if pipeline.scripts:
-                pipeline.load_scripts()
-
             # Run the pipeline
-            raw_results = self._execute_pipeline(args=script_args,
-                                                 client=client)
+            if self.can_replicate_commands:  # Redis 3.2 or higher
```
jkemp101: Are we sure after `redis.replicate_commands()` is called a `_G['f_' .. sha]()` call is turned back into a regular `EVALSHA` command when it is prepared to be sent over to the slave node?  Just want to make sure it isn't silently ignored.

tasktiger/redis_scripts.py
```diff
@@ -469,20 +480,39 @@ def execute_pipeline(self, pipeline, client=None):
...
+
+                raw_results = self._execute_pipeline(args=script_args,
+                                                     client=client)
+            else:
+                executing_pipeline = client.pipeline()
+
+                # Always load scripts to avoid issues when Redis loads data
+                # from AOF file / when replicating.
+                for s in pipeline.scripts:
+                    executing_pipeline.script_load(s.script)
```
jkemp101: We should monitor Redis CPU and network bandwidth after deploying this change to see if there is any noticeable impact.

tasktiger/redis_scripts.py
thomasst: Yes.

```
In [1]: s1 = app.redis.register_script('''redis.call('incr', 'x')''')

In [2]: s2 = app.redis.register_script('''redis.replicate_commands(); _G['f_{}']()'''.format(s1.sha))

In [3]: app.redis.script_load(s1.script)
Out[3]: '055c90c3af205164e578f6e1cb7ee167b810c6dd'

In [4]: app.redis.script_load(s2.script)
Out[4]: 'b7b28fdcb955f97bea62c60cc0f83be0f8ae6704'

# Restart slave here to empty its scripts cache and run MONITOR

In [5]: s2()

# Slave output:

MONITOR
+OK
+1515093425.210639 [0 127.0.0.1:6379] "SELECT" "0"
+1515093425.210681 [0 127.0.0.1:6379] "MULTI"
+1515093425.210689 [0 127.0.0.1:6379] "incr" "x"
+1515093425.210702 [0 127.0.0.1:6379] "EXEC"
```

PR #103, opened 2018-03-14
Task.tasks_from_queue: use tiger._key instead of hardcoded prefix/key

maldn:
when using the configurable prefix (i have multiple instances on one redis server) tasks_from_queue was searching for the wrong key, and the de-serialization of the data failed.


maldn:
just saw that you fixed this in #104

PR #105, opened 2018-03-27
Allow calling a function before/after task execution in the child process

thomasst:
Allow executing code in the child before/after task execution to do any needed initialization/clean up, e.g. initializing crypto modules.

Celery has signals/decorators for this, but I don't think this has to be fancy. Additional callbacks or kwargs can be added as needed.


thomasst:
Are you suggesting the context manager should be able to capture a task error? If so, it would have to be right around the task execution (within `_execute_forked`). I'm trying to make these functions independent from the task itself (they should be related to forking the child process), but if you think it's useful I can certainly make it that way. Do you have any specific use case? Custom error reporting?

Otherwise, if I insert it at the current place, the only exception it could capture is if both the task and `self.connection.rpush` fail, which isn't too useful.

tsx:
Custom error reporting is certainly an interesting case, although it is not the primary reason I'm suggesting to use context managers. If it is easy, let's do it, otherwise I wouldn't bother for now (we can move it inside `_execute_forked` without any impact on the API later if we decide to).

My thinking is more of a code interface perspective. Context manager *is* the python idiom for init/cleanup pattern. Suppose we want to open and close a database transaction for every. It is very likely that you already have a context manager that does the job for you, so all you have to do is pass it in there, instead of defining two brand new functions just for this. And it would be awkward to use pairs of functions if you need to share state between init/cleanup functions, while it is perfectly natural for class-based context managers

We can even expose context manager instance somehow to the running task (out of scope of this PR though).

thomasst:
@jkemp101 What exactly did we decide at the team retreat regarding this? Did we want just one setting that takes an array of context managers?

tsx:
👎 array of context managers. We should accept only one for API and implementation simplicity.

If anyone needs multiple context managers, they are free to compose them into one in their own setup code. They can use `https://docs.python.org/2/library/contextlib.html#contextlib.nested` (I know it's deprecated but the docs say you can still use it if you want variable number of ctx managers) or something similar.

jkemp101:
I'm not sure if we agreed on an implementation but the two potential use cases I have are:

* Flushing Prometheus metrics to disk at the end of every task.  My initial attempt to try this failed because it required importing app code in the config file to get at the function I wanted to be called.
* Calling SQLAlchemy's `Engine.dispose()` to cleanup DB connections before the task child process exits.

thomasst:
> contextlib.nested

Python 3 has this: https://docs.python.org/3.6/library/contextlib.html#contextlib.ExitStack

So wouldn't it be a common use case to have multiple independent context managers or enter/exit functions?

thomasst:
We should move forward on this since we're using this right now ourselves, and it makes it harder to use other TaskTiger improvements as long as this is pending.

Personally I don't see an issue with having either an array of context managers, or an array of functions.

alanhamlett:
👍 for an array of context managers as that would keep configs really clean. Otherwise signals work too and everyone is familiar with them from Flask.

jkemp101:
I'm fine with an array of context managers

jkemp101:
@thomasst Can you take a look at this now

Issue #112, opened 2018-07-28
TypeError when loading task json data


alanhamlett:
```
Traceback (most recent call last):
  File "./test.py", line 4, in purge_tasktiger_error_queues
    count, tasks = Task.tasks_from_queue(tiger, queue, ERROR)
  File "./venv/lib/python3.5/site-packages/tasktiger/task.py", line 393, in tasks_from_queue
    data = json.loads(serialized_data)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'NoneType'
```

thomasst:
What lead to this?

PR #114, opened 2018-08-14
Prevent TypeError when task json data missing

alanhamlett:
Related to #112.

This prevents an exception when calling `Task.tasks_from_queue`, but this doesn't fix the root problem of the task being created with invalid json data:

https://github.com/closeio/tasktiger/blob/10b5fe7057f43eb1dd5cbdc2035b1ab4c3cfc1d4/tasktiger/task.py#L314

### DIFF
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -3,10 +3,25 @@ import json
 import redis
 import time
 
-from ._internal import *
+from ._internal import (
+    ACTIVE,
+    QUEUED,
+    ERROR,
+    TaskImportError,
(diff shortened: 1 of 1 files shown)


thomasst:
> this doesn't fix the root problem of the task being created with invalid json data:

I've seen this sometimes, but usually only if I manually messed with Redis to clear keys and didn't properly clean up, in which case the key doesn't exist.

I thought of making this change as well, but I think I'd prefer us to fail explicitly here.

Any further thoughts @jkemp101 ?

PR #115, opened 2018-09-27
Use semaphore for queue lock

jkemp101:
- [x] Add tests for max workers > 1
- [ ] Test Redis impact with large number of locked queues, large number of workers, and new tasks being queued to locked queues

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 import importlib
 import logging
 import redis
+import time
 import structlog
 
 from .redis_scripts import RedisScripts
@@ -12,7 +13,7 @@ from .exceptions import *
 from .retry import *
 from .schedule import *
 from .task import Task
-from .worker import Worker
+from .worker import LOCK_REDIS_KEY, Worker
 
 __all__ = ['TaskTiger', 'Worker', 'Task',
 
@@ -353,6 +354,25 @@ class TaskTiger(object):
         """Get total queue size for QUEUED, SCHEDULED, and ACTIVE states."""
 
         return sum(self.get_queue_sizes(queue).values())
+
+    def queue_system_lock(self, queue, timeout):
+        """
+        Set system lock on a queue.
+
+        Max workers for this queue must be used for this to have any effect.
+
+        This will keep workers from processing tasks for this queue until
+        the timeout has expired. Active tasks will continue processing their
+        current task.
+
+        timeout is number of seconds to hold the lock
+        """
+
+        key = self._key(LOCK_REDIS_KEY, queue)
+        pipeline = self.connection.pipeline()
+        pipeline.zadd(key, 'SYSTEM_LOCK', time.time()+timeout)
+        pipeline.expire(key, timeout + 10)  # timeout plus buffer for troubleshooting
+        pipeline.execute()
 
     def get_queue_stats(self):
         """
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -150,6 +150,11 @@ class TaskTiger(object):
             # locking techniques.
             'SINGLE_WORKER_QUEUES': [],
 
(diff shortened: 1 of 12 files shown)


tasktiger/redis_semaphore.py
```diff
@@ -0,0 +1,35 @@
...
+        with open(os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                               'lua/semaphore.lua')) as f:
+            self._semaphore = self.redis.register_script(f.read())
+
+    def release(self):
+        """Release semaphore."""
+
+        self.redis.zrem(self.name, self.id)
+
+    def acquire(self):
```
tsx: We could add an optional `def acquire(self, time=time.time):` parameter, pass time explicitly in tests and avoid freezefrog dependency.

tasktiger/worker.py
```diff
@@ -353,24 +367,26 @@ def _get_queue_lock(self, queue, log):
...
+                max_workers = 1
                 break
 
         # Single worker queues require us to get a queue lock before
         # moving tasks
-        if single_worker_queue:
-            queue_lock = Lock(self.connection, self._key('qlock', queue),
-                              timeout=self.config['ACTIVE_TASK_UPDATE_TIMEOUT'])
-            acquired = queue_lock.acquire(blocking=False)
+        if max_workers:
```
thomasst: Update "Single worker queues" comment above.

tasktiger/__init__.py
```diff
@@ -150,6 +151,11 @@ def __init__(self, connection=None, config=None, setup_structlog=False):
             # locking techniques.
             'SINGLE_WORKER_QUEUES': [],
 
+            # The maximum number of workers that will be allowed to actively
+            # process tasks for any queue. Similar to Single Worker Queues
+            # but allows specifying any number 1 or greater.
+            'MAX_WORKERS_PER_QUEUE': None,
```
thomasst: Should there be  an option to configure this per queue, and then deprecate `SINGLE_WORKER_QUEUES`? E.g.:
```
MAX_WORKERS_PER_QUEUE = {
    None: 4,    # default
    'some-queue': 1,
    'some-queue.subqueue': 2,
    'other-queue': 6,
}
```

tasktiger/__init__.py
```diff
@@ -323,6 +331,25 @@ def delay(self, func, args=None, kwargs=None, queue=None,
 
         return task
 
+    def queue_system_lock(self, queue, timeout):
```
thomasst: Is this used anywhere?

tasktiger/worker.py
```diff
@@ -11,16 +11,20 @@
...
 
 from ._internal import *
 from .exceptions import RetryException, TaskNotFound
+from .redis_semaphore import Semaphore
 from .retry import *
 from .stats import StatsThread
 from .task import Task
 from .timeouts import UnixSignalDeathPenalty, JobTimeoutException
 
+LOCK_REDIS_KEY = 'qlock'
```
thomasst: Should consider adding this to the "Redis keys" comment section in `tasktiger/__init__.py`.

tasktiger/__init__.py
jkemp101: I'm a bit torn on how we should handle queue configuration as we add more and more queue specific options. Right now a queue's configuration is spread across multiple keys in the config. Do you think we should consider moving to something like this instead and stop using `ONLY_QUEUES`, `BATCH_QUEUES`, `SINGLE_WORKER_QUEUES`/`MAX_WORKERS_PER_QUEUE`?
```
QUEUES = {
    'a': {
        max_workers: 5,
        batch: 50
    },
    'b': {}
}
```
We could also make it smart enough to detect if `ONLY_QUEUES` is a list or a dictionary so we don't have to add `QUEUES` to the config. 

If we wanted to make that change I could remove `MAX_WORKERS_PER_QUEUE` for now and just support it as a command line option until we refactor the config handling.

PR #116, opened 2018-10-09
Redis migrate script

jkemp101:
Utility script to copy tasktiger keys from one Redis instance to another

### DIFF
--- a/scripts/move_tasks.py
+++ b/scripts/move_tasks.py
@@ -11,39 +11,38 @@ PREFIX = 't'
 
 def move_tasks(source, dest, state, queue, apply):
     """Move tasks in a queue."""
-    key = '{}:{}:{}'.format(PREFIX, state, queue)
-    tasks = source.zrange(key, 0, -1)
-    dump = source.dump(key)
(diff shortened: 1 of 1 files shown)


jkemp101:
I think it fails on all versions of Redis but in 3.0+ you can use the REPLACE option to overwrite it.  Assuming you are moving to a fresh Redis DB I would carefully use FLUSHDB on the destination db before rerunning the script.  We could add support for the REPLACE option but I think in most cases if the key already exists something is really wrong and you might need to abort the migration anyway.

scripts/move_tasks.py
```diff
@@ -55,7 +54,7 @@ def run(apply=False):
     move_queues_in_state(source, dest, 'queued', apply)
     move_queues_in_state(source, dest, 'active', apply)
     move_queues_in_state(source, dest, 'scheduled', apply)
-    # move_queues_in_state(source, dest, 'error', apply)
+    move_queues_in_state(source, dest, 'error', apply)
```
thomasst: Are we going to move error queues?

PR #117, opened 2018-11-05
Lazy tasktiger initialization

tsx:
This lets you set up connection and config later during initialization so you don't have to fiddle with import ordering.


tsx:
Just for the record: I went with explicit `lazy_init` param because `connection=None` and `config=None` is a real working configuration: it connects to localhost redis and provides some sensible defaults. Since it's an open project and people might depend on this configuration, I decided not to break existing behavior and go with an extra param.

PR #118, opened 2018-11-06
Support max queue size

jkemp101:
- [ ] Breakout the one test and add more

### DIFF
--- a/tasktiger/stats.py
+++ b/tasktiger/stats.py
@@ -1,7 +1,9 @@
 import threading
 import time
 
-from ._internal import g_fork_lock
+import redis
+
+from ._internal import ACTIVE, g_fork_lock, QUEUED, SCHEDULED
 
 class StatsThread(threading.Thread):
     def __init__(self, tiger):
@@ -65,3 +67,13 @@ class StatsThread(threading.Thread):
 
     def stop(self):
         self._stop_event.set()
+
+
+def get_queue_size(tiger, queue_name):
+    """Get size of this task's queue for each state."""
+    pipeline = tiger.connection.pipeline()
+    pipeline.zcard(tiger._key(QUEUED, queue_name))
+    pipeline.zcard(tiger._key(SCHEDULED, queue_name))
+    pipeline.zcard(tiger._key(ACTIVE, queue_name))
+    results = pipeline.execute()
+    return dict(zip([QUEUED, SCHEDULED, ACTIVE], results))
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -4,7 +4,8 @@ import redis
 import time
 
 from ._internal import *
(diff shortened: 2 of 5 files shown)


tasktiger/stats.py
```diff
@@ -65,3 +67,13 @@ def run(self):
 
     def stop(self):
         self._stop_event.set()
+
+
+def get_queue_size(tiger, queue_name):
+    """Get size of this task's queue for each state."""
+    pipeline = tiger.connection.pipeline()
+    pipeline.zcard(tiger._key(QUEUED, queue_name))
```
tsx:
```python
stats = [QUEUED, SCHEDULED, ACTIVE]
for stat in stats:
   pipeline.zcard(..., stat, ...)
...
return ...zip(stats)...
```

tasktiger/task.py
```diff
@@ -290,12 +291,19 @@ def delay(self, when=None):
         if tiger.config['ALWAYS_EAGER'] and state == QUEUED:
             return self.execute()
 
+        if max_queue_size:
+            # This will fail adding a unique task that already is queued but
+            # the queue size is at the max
+            queue_size = sum(get_queue_size(self.tiger, self.queue).values())
```
tsx: The `sum` thing deserves a separate method `tiger.total_queue_size(queue)`

tasktiger/task.py
```diff
@@ -290,12 +291,19 @@ def delay(self, when=None):
         if tiger.config['ALWAYS_EAGER'] and state == QUEUED:
             return self.execute()
 
+        if max_queue_size:
```
tsx: This needs to be moved up at least above eager execution.

tasktiger/task.py
tsx: There's also a local `tiger` already set.

tests/test_base.py
```diff
@@ -1047,3 +1049,68 @@ def test_single_worker_queue(self):
         self._ensure_queues()
 
         worker.join()
+
+
+class TestMaxQueue(BaseTestCase):
```
tsx: I'd prefer a separate file: this one is already too long (>1k lines).

tasktiger/task.py
```diff
@@ -287,6 +296,13 @@ def delay(self, when=None):
         # ensure there are no serialization errors.
         serialized_task = json.dumps(self._data)
 
+        if max_queue_size:
+            # This will fail adding a unique task that already is queued but
+            # the queue size is at the max
+            queue_size = tiger.get_total_queue_size(self.queue)
```
thomasst: I'm late to this one, but it would be nice if this was done as part of the pipeline. E.g. abort the pipeline if the queue is full. Rather than doing a round trip before queueing each task, which is also prone to race conditions (though it's no big deal).

PR #125, opened 2019-01-16
Refactor system lock

jkemp101:
* Moved system lock logic into Semaphore class
* Changes zset score to be the time the lock expires versus when it was acquired. Will merge https://github.com/closeio/tasktiger/pull/124/files with this change so there aren't any backwards compatibility issues.
* Don't use `max` and `id` as variable names
* Add more docs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -673,6 +673,16 @@ executed.
           print(task.n_executions())
 
 
+Pause queue processing
+----------------------
+
+The ``--max-workers-per-queue`` option uses queue locks to control the
+number of workers that can simultaneously process the same queue. When using
+this option a system lock can be placed on a queue which will keep workers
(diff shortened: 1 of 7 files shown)


jkemp101:
I'm still not sure what the best expiration is for these keys. It is only for diagnostic purposes and isn't supposed to have any affect on the locking mechanism. I purposely used `+ 10` for the system lock because I could image very long lock times like 1 year which would then have very long expiration times if I `* 2`. On the other hand I imagined shorter lock times for normal usage so I wanted the expiration to be more proportional to the timeouts.  In some high volume locking scenarios the `* 2` might even end up leaving too many keys around. We may end up just expiring at the timeout time once it is tried and true.

PR #127, opened 2019-01-28
Limit execution histories and structlog processor

jkemp101:
Adds:
* New option to disable saving tracebacks in execution histories
* Structlog processor that automatically injects task id

### DIFF
--- a/tasktiger/logging.py
+++ b/tasktiger/logging.py
@@ -9,6 +9,6 @@ def tasktiger_processor(logger, method_name, event_dict):
     """
 
     if g['current_tasks'] is not None and not g['current_task_is_batch']:
-        event_dict['task_id'] = self.tiger.current_task.id
+        event_dict['task_id'] = g['current_tasks'][0].id
 
     return event_dict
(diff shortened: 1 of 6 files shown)


tasktiger/logging.py
```diff
@@ -0,0 +1,14 @@
...
+
+def tasktiger_processor(logger, method_name, event_dict):
+    """
+    TaskTiger structlog processor.
+
+    Inject the current task id for non-batch tasks.
+    """
+
+    if g['current_tasks'] is not None and not g['current_task_is_batch']:
+        event_dict['task_id'] = g['current_tasks'][0].id
```
caseywebdev: Is it possible for `g['current_tasks']` to be `[]` here? Or is that value either a `None` or a non-empty list?

jkemp101: I kind of based that on [this code](https://github.com/closeio/tasktiger/blob/fda9d01e9eb5e31508a1420a3d11367f85c4c2f2/tasktiger/__init__.py#L208) which looks like we already assume `g['current_tasks']` won't be `[]`.

PR #129, opened 2019-03-06
Fix pinned packages

jkemp101:


jkemp101:
Yep, I agree with all of that. This at least gets us back to be closer to what we had before I pinned specific versions.

PR #130, opened 2019-05-22
Clarify what happens when requeueing a unique task

jkemp101:
Does this clarification make sense?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -358,7 +358,7 @@ The following options can be only specified in the task decorator:
 
   The schedule function must return the next task execution datetime, or
   ``None`` to prevent periodic execution. The function is executed to determine
-  the initial tasks execution date when a worker is initialized, and to determine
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -357,7 +358,7 @@ The following options can be only specified in the task decorator:
 
   The schedule function must return the next task execution datetime, or
   ``None`` to prevent periodic execution. The function is executed to determine
-  the initial taks execution date when a worker is initialized, and to determine
+  the initial tasks execution date when a worker is initialized, and to determine
```
thomasst: *task

"tasks" would be grammatically incorrect. "task's" is also incorrect since it implies that there are multiple tasks.

jkemp101: Fixed. I probably should read the sentence next time, I just assumed a `s` was dropped not `sk` was reversed.

PR #135, opened 2019-06-24
Remove lock when queueing periodic tasks

jkemp101:
@tsx pointed out that my last PR https://github.com/closeio/tasktiger/pull/132 has a regression since we use a single lock to queue periodic tasks we can no longer rely on `(assuming other process will)` since they may be queueing tasks for queues the current process doesn't process (or vice versa). Potential fixes:
* Remove lock - not sure if it is really needed
* Revert previous PR
* Implement a more complicated per task or per queue locking mechanism

@thomasst do you remember what the lock was doing for this?

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -146,9 +146,6 @@ class TaskTiger(object):
             # How often to print stats.
             'STATS_INTERVAL': 60,
 
-            # Upper bound for the time it takes to queue all periodic tasks.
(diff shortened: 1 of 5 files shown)


thomasst:
Good point that I missed :(

I think the worst thing that could happen is that we'd call `task._queue_for_next_period()` multiple times. Since they're unique I don't see an issue with just dropping the lock. I think the main point was to prevent multiple workers from doing duplicate work if starting them around the same time.

PR #139, opened 2019-08-09
Allow custom context managers to see task errors

igor47:
we use user-specified context managers to wrap `self._execute_forked` but that
function always catches all exceptions. this means that our `__exit__` function
always thinks the tasks raised no error, and so do not necesserily perform
proper cleanup.
   
here, we move the context-manager wrapping closer to where we invoke the
function. the downside is that, if we swallow exceptions in the context
manager, we won't get retry logic, which i think is an acceptable tradeoff

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -314,32 +314,33 @@ class Worker(object):
             is_batch_func = getattr(func, '_task_batch', False)
             g['current_task_is_batch'] = is_batch_func
 
-            if is_batch_func:
-                # Batch process if the task supports it.
-                params = [{
-                    'args': task.args,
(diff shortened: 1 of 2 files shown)


thomasst:
> if we swallow exceptions in the context
> manager, we won't get retry logic, which i think is an acceptable tradeoff

This would only be an issue if we explicitly return `True` in the `__exit__` method of the context manager, right?

igor47:
> This would only be an issue if we explicitly return `True` in the `__exit__` method of the context manager, right?

yes, that's correct

thomasst:
Thanks! Fixed conflicts & reformatted in https://github.com/closeio/tasktiger/pull/165

Issue #20, opened 2016-07-13
Log task's execution time


wojcikstefan:
Right now it's very hard to scan the logs and determine how much time was spent executing a given task.

Add `duration` to these log statements. I think `now()` - `task.ts` should be close enough since `_ts` is set [here](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/worker.py#L596-L597). Might need a little investigation.
* [done](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/worker.py#L770)
* [task error](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/worker.py#L850)

PR #142, opened 2019-09-05
Log Task Approximate Execution Time

AlecRosenbaum:
Per https://github.com/closeio/tasktiger/issues/20, the `task.ts` here should be _close enough_ that the information will still be useful.

In cases where the batch size is greater than one, this won't be super accurate. `task.ts` is set based on when the batch starts, so if the batch size is `50`, then later tasks will report cumulative execution time for the batch rather than just the individual task.

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -764,10 +764,13 @@ class Worker(object):
         """
         log = self.log.bind(queue=queue, task_id=task.id)
 
+        when = time.time()
+        execution_duration = task.ts and when - task.ts
+
         def _mark_done():
             # Remove the task from active queue
             task._move(from_state=ACTIVE)
-            log.info('done')
+            log.info('done', execution_duration=execution_duration)
 
         if success:
             _mark_done()
@@ -809,10 +812,9 @@ class Worker(object):
 
             state = ERROR
 
-            when = time.time()
-
             log_context = {
-                'func': task.serialized_func
+                'func': task.serialized_func,
+                'execution_duration': execution_duration,
             }
 
             if should_retry:
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -627,7 +627,7 @@ class Worker(object):
             log.debug('processed', attempted=len(tasks),
                       processed=processed_count)
             for task in processed_tasks:
-                self._finish_task_processing(queue, task, success)
+                self._finish_task_processing(queue, task, success, now)
 
         return processed_count
 
@@ -756,7 +756,7 @@ class Worker(object):
 
         return success, ready_tasks
 
-    def _finish_task_processing(self, queue, task, success):
+    def _finish_task_processing(self, queue, task, success, start_time):
         """
         After a task is executed, this method is called and ensures that
         the task gets properly removed from the ACTIVE queue and, in case of an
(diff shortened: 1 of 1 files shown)


tasktiger/worker.py
```diff
@@ -764,10 +764,13 @@ def _finish_task_processing(self, queue, task, success):
...
         log = self.log.bind(queue=queue, task_id=task.id)
 
+        when = time.time()
+        execution_duration = task.ts and when - task.ts
+
         def _mark_done():
             # Remove the task from active queue
             task._move(from_state=ACTIVE)
-            log.info('done')
+            log.info('done', execution_duration=execution_duration)
```
thomasst: I'm sure there might be an even better name, but would `processing_duration` be more accurate? I'd expect `execution_duration` to be the duration that it takes to actually execute the task, i.e. if it was measured somewhere within `_execute`, and not what we're actually doing here.

AlecRosenbaum:
I was doing some manual testing to double check that this works correctly and saw some odd results. I think I tracked down the source of the issue, but I'm not sure what the best course forward is.

When handling periodic tasks, [`_execute`](https://github.com/closeio/tasktiger/blob/613bd875f506b2119633f524f8aa6dde2b99d154/tasktiger/worker.py#L624) schedules the next task before executing the current task by calling [`task._queue_for_next_period`](https://github.com/closeio/tasktiger/blob/613bd875f506b2119633f524f8aa6dde2b99d154/tasktiger/worker.py#L438). `task._queue_for_next_period` calls [`self.delay(when=when)`](https://github.com/closeio/tasktiger/blob/613bd875f506b2119633f524f8aa6dde2b99d154/tasktiger/task.py#L477). `self.delay` ends up setting [`self._ts`](https://github.com/closeio/tasktiger/blob/613bd875f506b2119633f524f8aa6dde2b99d154/tasktiger/task.py#L320) to the future execution time, which overwrites the original `_ts` value set when scheduling the task that's being executed. Of course this ends up causing funny values when reporting processing duration because the calculation ends up being `now - next_scheduled_time` instead of `now - this_tasks_scheduled_time`.

So it seems like for periodic tasks we're inadvertently mixing state of the current task execution with state of the next task execution. My inclination would be to change `_queue_for_next_period` to clone the task (with something like `Task(_data=self._data, _func=self._func)`) and return the new task clone after scheduling. `_queue_for_next_period` seems to only be used in two places, which would be easy to convert to use the new call signature.

Thoughts?

AlecRosenbaum:
Per conversation in slack, the easiest approach was just changing the call signature of `_finish_task_processing` and avoiding `task.ts` entirely.

I've done some manual testing on this and the reported `processing_duration`'s seem much more realistic.

tasktiger/worker.py
```diff
@@ -756,18 +756,21 @@ def _execute_task_group(self, queue, tasks, all_task_ids, queue_lock):
...
-    def _finish_task_processing(self, queue, task, success):
+    def _finish_task_processing(self, queue, task, success, start_time):
         """
         After a task is executed, this method is called and ensures that
         the task gets properly removed from the ACTIVE queue and, in case of an
         error, retried or marked as failed.
         """
         log = self.log.bind(queue=queue, task_id=task.id)
 
+        when = time.time()
```
thomasst: I'd name it `end_time` or `now`. I think I used `when` sometimes because I'd pass it into `delay(when=when)`, but I don't think it makes any sense in this scenario.

PR #145, opened 2019-09-27
Apply and enforce black formatting

AlecRosenbaum:
This change applies and enforces code formatting with the [`black`](https://github.com/psf/black) tool.

The formatting settings are consistent with the settings used in other `closeio` projects.

### DIFF
--- a/tasktiger/__init__.py
+++ b/tasktiger/__init__.py
@@ -487,32 +487,32 @@ class TaskTiger(object):
 @click.option(
     '-q',
     '--queues',
-    help='If specified, only the given queue(s) '
-    'are processed. Multiple queues can be '
-    'separated by comma.',
+    help=(
+        'If specified, only the given queue(s) are processed. Multiple queues '
+        'can be separated by comma.'
+    ),
 )
 @click.option(
     '-m',
     '--module',
-    help="Module(s) to import when launching the "
-    "worker. This improves task performance "
-    "since the module doesn't have to be "
-    "reimported every time a task is forked. "
-    "Multiple modules can be separated by "
-    "comma.",
+    help=(
+        'Module(s) to import when launching the worker. This improves task '
+        'performance since the module doesn\'t have to be reimported every '
+        'time a task is forked. Multiple modules can be separated by comma.'
+    ),
 )
 @click.option(
     '-e',
     '--exclude-queues',
-    help='If specified, exclude the given '
-    'queue(s) from processing. '
-    'Multiple queues can be '
-    'separated by comma.',
+    help=(
+        'If specified, exclude the given queue(s) from processing. Multiple '
+        'queues can be separated by comma.'
+    ),
 )
 @click.option(
     '-M',
     '--max-workers-per-queue',
-    help='Maximum workers allowed ' 'to process a queue',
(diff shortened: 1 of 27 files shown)


tasktiger/__init__.py
```diff
@@ -449,34 +482,56 @@ def get_queue_stats(self):
...
-                                             'separated by comma.')
-@click.option('-M', '--max-workers-per-queue', help='Maximum workers allowed '
-                                                    'to process a queue', type=int)
-@click.option('--store-tracebacks/--no-store-tracebacks',
-              help='Store tracebacks with execution history',
-              default=None)
+@click.option(
+    '-q',
+    '--queues',
+    help='If specified, only the given queue(s) '
```
thomasst: These strings can be longer now.

tasktiger/_internal.py
```diff
@@ -47,49 +44,57 @@ def import_attribute(name):
...
 
 def serialize_func_name(func):
     """
     Returns the dotted serialized path to the passed function.
     """
     if func.__module__ == '__main__':
-        raise ValueError('Functions from the __main__ module cannot be '
-                         'processed by workers.')
+        raise ValueError(
+            'Functions from the __main__ module cannot be '
```
thomasst: String can be longer

tasktiger/redis_scripts.py
```diff
@@ -388,23 +415,29 @@ def zpoppush(self, source, destination, count, score, new_score,
 
                 if not on_success or on_success[0] != 'update_sets':
                     raise NotImplementedError()
-                set_value, remove_from_set, add_to_set, add_to_set_if_exists \
-                    = on_success[1:]
+                set_value, remove_from_set, add_to_set, add_to_set_if_exists = on_success[
```
thomasst: Is there a way to format this nicer? It's over the line length.

tasktiger/__init__.py
```diff
@@ -116,56 +126,45 @@ def init(self, connection=None, config=None, setup_structlog=False):
         self.config = {
             # String that is used to prefix all Redis keys
             'REDIS_PREFIX': 't',
-
```
thomasst: No strong opinion here but it was more readable with the newlines.

tasktiger/worker.py
```diff
@@ -196,29 +214,34 @@ def _wait_for_new_tasks(self, timeout=0, batch_timeout=0):
...
+                        timeout,
+                        batch_timeout,
                     )
 
                 message = self._pubsub.get_message()
 
             if self._did_work:
-                break   # Exit immediately if we did work during the last
-                        # execution loop because there might be more work to do
+                break  # Exit immediately if we did work during the last
```
thomasst: Reformat these comments to be on top now.

tasktiger/__init__.py
AlecRosenbaum: I'm inclined to agree, but I think the only way to keep them is by creating an unformatted block. I'd rather not create unformatted blocks unless there's a breaking change though.

jkemp101: Looks like Black would allow a plain `#` to add a blank line between each setting if we think that helps readability. Your call if that helps or hurts.

Issue #146, opened 2019-10-01
Periodic tasks can get queued with different default args/kargs


jkemp101:
When tasktiger starts and automatically schedules periodic tasks the task ID is the hash of:
```{"args": null, "func": "my_function", "kwargs": null}```

If the same task is scheduled in a Python shell with `my_function.delay()` the hash is of:
```{"args": [], "func": "my_function", "kwargs": {}}```
which will be a different ID than the previous version. This can cause the same periodic task to be scheduled twice with two different IDs.

`[]` and `{}` are probably the more sane defaults. Two options to fix:
* Force the hash to be `None` for `[]` and `{}` in [`gen_unique_id`](https://github.com/closeio/tasktiger/blob/ad744574c053ed7cba3f10cabea3770c6c2e57f6/tasktiger/_internal.py#L56)
* Calculate the hash using `[]` and `{}` and have a migration to unschedule existing tasks that can be used before deploying the new version of TaskTiger so that two sets of periodic tasks don't end up scheduled.

PR #147, opened 2019-10-02
Fix Default Periodic Task Unique ID

AlecRosenbaum:
Related Issue: https://github.com/closeio/tasktiger/issues/146

Per the issue, manually scheduled and automatically queued period tasks _should_ have the same unique id, but currently don't.
```python
periodic_task.delay()  # id == hash(function_name, [], {})
Worker().run()  # periodic_task.id == hash(function_name, None, None)
```

If this happens then there can end up being multiple versions of an argument-less periodic task floating around in the queue.

This change fixes that behavior by always defaulting args/kwargs to `[]`/`{}` before generating the unique id. Also, this change introduces behavior to self-correct ids when rescheduling the task for the future. I've added a test confirming this behavior, but it is a departure from what the current behavior is. I'm not sure if we want to keep that aspect of this change or just say that tasks should be manually identified/purged when there's an issue like the one linked.

### DIFF
--- a/tests/test_base.py
+++ b/tests/test_base.py
@@ -56,10 +56,21 @@ class BaseTestCase:
     def _ensure_queues(
         self, queued=None, active=None, error=None, scheduled=None
     ):
+
+        expected_queues = {
+            'queued': {name for name, n in (queued or {}).items() if n},
+            'active': {name for name, n in (active or {}).items() if n},
+            'error': {name for name, n in (error or {}).items() if n},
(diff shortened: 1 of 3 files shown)


tests/test_base.py
```diff
@@ -56,10 +56,21 @@ def teardown_method(self, method):
     def _ensure_queues(
         self, queued=None, active=None, error=None, scheduled=None
     ):
+
```
AlecRosenbaum: These lines aren't related to the other functionality in this PR, but the improved error message helped me a bit while debugging tests.

Issue #137, opened 2019-07-24
Add functionality to cleanup errored tasks


jkemp101:
Failed tasks can be left in the `ERROR` state and must be manually removed. There should be an easy way to purge old errored tasks.

* Add method `purge_errored_tasks(queues=None, exclude_queues=None, last_execution_before=None)` to the [TaskTiger](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/__init__.py#L77) class
   * `queues` and `exclude_queues` should function similar to the way the [Worker](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/worker.py#L51) class processes them, using config values if they aren't specified in the function call
   * `last_execution_before` seconds since the epoch that can be used to only delete errored tasks older than the specified age OR we could specify age based on last execution in seconds (e.g. tasks last executed more than 12 weeks ago = 7,257,600 seconds)
* Once tasks are identified their [delete()](https://github.com/closeio/tasktiger/blob/a93b4fe03ec807dc82df4504e3b62910848c3e00/tasktiger/task.py#L458) method can be called to remove them.
* Might want to include a way to slow down the deletes in case there are millions to process and the caller doesn't want to impact Redis

AlecRosenbaum:
> Might want to include a way to slow down the deletes in case there are millions to process and the caller doesn't want to impact Redis

Will the throttled call be the primary use case? My initial thought is to have `purge_errored_tasks` return a iterator where each item yielded represents processing `batch_size` items, but if throttled requests isn't the primary usage pattern I might want to do something different. Example usage:
```python
# process all
all(purge_errored_tasks(...))

# throttle batches
for n_processed in purge_errored_tasks(...):
    print('Progress: {n} error\'d tasks have been purged so far')
    time.sleep(.1)
print('Done')
```
Then the `time.sleep` call could be easily replaced with some sort smarter throttle using something like `limitlion`, based on current redis usage metrics, etc.

AlecRosenbaum:
Per some offline discussion, I'll be updating the call signature from `batch_size` to `limit`, which makes the call pattern much better and allows it to return the number of deleted tasks rather than an iterator.

New example usage:
```python
# processing all
limit = 10000
n_processed = limit
while n_processed == limit:
    n_processed = purge_errored_tasks(..., limit=limit)
```

PR #148, opened 2019-10-02
Add functionality to purge errored tasks

AlecRosenbaum:
Related Issue: https://github.com/closeio/tasktiger/issues/137

This change adds a `TaskTiger.purge_errored_tasks` method that makes it much easier to (surprise, surprise) purge errored tasks. Arguments allow the caller to both specify specific queues, and specify limits as to not overload redis in one method call.

Example method call:
```python
limit = 10000
n_processed = limit
while n_processed == limit:
    n_processed = tiger.purge_errored_tasks(queues=['default'], limit=limit)
    time.sleep(.5)
```


tasktiger/__init__.py
```diff
@@ -482,6 +485,108 @@ def get_queue_stats(self):
...
+
+        def errored_tasks():
+            queues_with_errors = self.connection.smembers(self._key(ERROR))
+            for queue in queues_with_errors:
+                if not match(queue):
+                    continue
+
+                skip = 0
+                total = None
+                task_limit = limit or 1000
```
jkemp101: This seems a little weird setting a limit of `0` would end up being `1000`. Maybe just assert `limit > 0`?

AlecRosenbaum: I've added an additional `assert` with the other argument validation

tasktiger/tasktiger.py
```diff
@@ -466,6 +475,92 @@ def get_queue_stats(self):
...
+                    for task in tasks:
+                        if (
+                            last_execution_before
+                            and task.ts
+                            and task.ts > last_execution_before
+                        ):
+                            continue
+                        yield task
+                    skip += task_limit
+                    total = total_tasks
```
jkemp101: I haven't tested this to confirm but it looks like the overall limit is being applied to the number of tasks we look at versus the number of tasks we delete. For instance, if we had 1001 total errored tasks and the first 1000's `task.ts` was greater than `last_execution_before` but the 1001 task wasn't, I don't think we would delete the 1001 task if `purge_errored_tasks()` was called with `limit=1000`.

README.rst
```diff
@@ -720,6 +720,23 @@ reported to Rollbar. Here is a custom worker launch script:
...
+this up as a periodic task as follows:
+
+..code:: python
+  from tasktiger import TaskTiger, periodic
+
+  tiger = TaskTiger()
+
+  @tiger.task(schedule=periodic(hours=1))
+  def purge_errored_tasks():
+      tiger.purge_errored_tasks(limit=1000)
```
jkemp101: I think I would add `last_execution_before` to the example within something like older than 3 months.

PR #150, opened 2019-10-10
Fix global logging namespace pollution

AlecRosenbaum:
While working on `tasktiger-admin`, I started getting some odd errors coming from `tasktiger` about `logging`.

It looks like somehow the `logging` module from `import logging` is getting overwritten with `tasktiger.logging` module, which then causes further issues when accessing `logging.DEBUG` and setting up `structlog`.

I don't know how this is happening, and I can't actually find anywhere that imports `tasktiger.logging`. I initially thought this might be a relative import issue, but adding `__future__` imports doesn't seem to help at all. When `logging` is initially imported, the module is correct. Somewhere between the initial import and actually creating the `TaskTiger` instance the global `logging` module gets swapped for `tasktiger.logging`.

The only other thing I can think of is if there's something funky that goes on with `__init__.py` files that overwrites global modules at the end of bootstrapping so that doing something like `import tasktiger.logging` actually resolves to `tasktiger/logging.py` instead of the `logging` module imported within the `tasktiger/__init__.py` file. Moving the core `TaskTiger` class into its own file seems to resolve the issue.


tasktiger/__init__.py
```diff
@@ -34,503 +18,13 @@
...
-@click.option('-a', '--password', help='Redis password')
-@click.option('-n', '--db', help='Redis database number')
-@click.pass_context
-def run_worker(context, host, port, db, password, **kwargs):
-    conn = redis.Redis(
-        host, int(port or 6379), int(db or 0), password, decode_responses=True
-    )
-    tiger = context.obj or TaskTiger(setup_structlog=True, connection=conn)
-    tiger.run_worker(**kwargs)
+from ._internal import *
```
mpessas: Imports should go above `__all__`.

AlecRosenbaum: Not according to [PEP8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names). I've actually had an issue open in `isort` for quite a while now https://github.com/timothycrosley/isort/issues/711

PR #154, opened 2019-12-10
Improve worker shutdown

jkemp101:
Once the parent process receives a SIGTERM the `select.select()` will always immediately return. This causes `self._heartbeat()` to be called repeatedly as fast as possible causing a large load on Redis until the child process exits.

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -1,4 +1,3 @@
-import binascii
 from collections import OrderedDict
 import errno
 import fcntl
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -598,9 +598,13 @@ class Worker(object):
                     if results[0]:
                         # Purge pipe so select will pause on next call
                         try:
+                            # Linux:
+                            #   Raises IOError if it would block
+                            #
+                            # macOS:
+                            #   Returns empty string if it would block
                             opened_fd.read(1)
                         except IOError:
-                            # Raised if read would block
                             pass
 
                 except select.error as e:
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -588,15 +588,20 @@ class Worker(object):
                     # macOS:
                     #   - 15 (SIGTERM) when parent receives SIGTERM
                     #   - 20 (SIGCHLD) when child exits
-                    select.select(
+                    results = select.select(
                         [pipe_r],
                         [],
                         [],
                         self.config['ACTIVE_TASK_UPDATE_TIMER'],
                     )
 
-                    # Purge pipe so select will pause on next call
-                    opened_fd.read(1)
+                    if results[0]:
+                        # Purge pipe so select will pause on next call
+                        try:
+                            opened_fd.read(1)
+                        except IOError:
+                            # Raised if read would block
+                            pass
 
                 except select.error as e:
                     if e.args[0] != errno.EINTR:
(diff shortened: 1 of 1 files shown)


tasktiger/worker.py
```diff
@@ -569,12 +570,30 @@ def check_child_exit():
...
                         [pipe_r],
                         [],
                         [],
                         self.config['ACTIVE_TASK_UPDATE_TIMER'],
                     )
+
+                    if result[0]:
+                        # Purge pipe so select will pause on next call.
+                        # This is a blocking call but the select should
+                        # ensure at least one byte is available.
```
thomasst: Sounds like this may not be guaranteed.

> Under Linux, select() may report a socket file descriptor as "ready for reading", while nevertheless a subsequent read blocks. This could for example happen when data has arrived but upon examination has wrong checksum and is discarded. **There may be other circumstances in which a file descriptor is spuriously reported as ready.** Thus it may be safer to use O_NONBLOCK on sockets that should not block.

http://man7.org/linux/man-pages/man2/select.2.html

tasktiger/worker.py
```diff
@@ -569,12 +575,34 @@ def check_child_exit():
...
                         [pipe_r],
                         [],
                         [],
                         self.config['ACTIVE_TASK_UPDATE_TIMER'],
                     )
+
+                    if results[0]:
+                        # Purge pipe so select will pause on next call
+                        try:
+                            opened_fd.read(1)
```
thomasst: I tried and it would just return an empty string, rather than raise an `IOError`. Obviously the code is still going to work as is even if we catch the error.

```
In [1]: import os                                                                                                                                                                                                 
                                                                                                                                                                                                                  
In [2]: pipe_r, pipe_w = os.pipe()

In [3]: opened_fd = os.fdopen(pipe_r)

In [5]: import fcntl

In [6]: flags = fcntl.fcntl(pipe_r, fcntl.F_GETFL, 0)
   ...: 

In [7]: flags = flags | os.O_NONBLOCK
   ...: 

In [8]: fcntl.fcntl(pipe_r, fcntl.F_SETFL, flags)
Out[8]: 0

In [9]: opened_fd.read(1)
Out[9]: ''

In [10]: opened_fd.read(1)
Out[10]: ''
```

jkemp101: Linux raises the exception, Mac doesn't.  I added a note since I explain the signal behavior above for both platforms and this is relevant also.

```
In [1]: import os

In [2]: pipe_r, pipe_w = os.pipe()

In [3]: opened_fd = os.fdopen(pipe_r)

In [4]: import fcntl

In [5]: flags = fcntl.fcntl(pipe_r, fcntl.F_GETFL, 0)

In [6]: flags = flags | os.O_NONBLOCK

In [7]: fcntl.fcntl(pipe_r, fcntl.F_SETFL, flags)
Out[7]: 0

In [8]: opened_fd.read(1)
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
/home/closeio/closeio/closeio/exporter/documents/__init__.py in <module>()
----> 1 opened_fd.read(1)

IOError: [Errno 11] Resource temporarily unavailable

In [9]:
```

jkemp101: Turns out Python 2.7 and 3.x behavior is different in Linux. 3.x doesn't raise an exception probably due to [PyErr_Clear()](https://github.com/python/cpython/blob/3.8/Modules/_io/fileio.c#L815) in Py3 that doesn't exist in the [2.7 implementation](https://github.com/python/cpython/blob/2.7/Modules/_io/fileio.c#L694-L695)?

Issue #156, opened 2019-12-17
is there any plan to support the Windows platform ?


thomasleveil:
Hi, 

In its current state (v0.10.1), this project is not compatible with the Windows platform due to the use of the fcntl module.

Is there any plan to make this project compatible with Windows ?

Also I guess the package descriptions should be fixed to reflect that fact. Currenty setup.py reads:

```python
setup(
    name='tasktiger',
    version='0.10.1',
    ...
    platforms='any',
    ...
    classifiers=[
        'Operating System :: OS Independent',
```

AlecRosenbaum:
I am not aware of any plans to support the windows platform. AFAIK, none of the maintainers run windows so we wouldn't catch any issues. There's a good amount of logic around process management that would probably not work as-is. Changes are typically just tested on Linux and OS X.

I didn't realize the package description indicated windows support, thank you for pointing it out. I'll fix it in with the next release.

cc/ @jkemp101 @thomasst

thomasleveil:
Thank you for that prompt reply. Should we let this opened until the package description get fixed?

AlecRosenbaum:
Yep! I'll close it with the PR that fixed it.

AlecRosenbaum:
`master` has the fix, it'll show up on pypi in the next release.

PR #157, opened 2019-12-17
Fix classifiers in setup.py

AlecRosenbaum:
Closes: https://github.com/closeio/tasktiger/issues/156

The classifiers in `setup.py` aren't correct, since they indicate windows support. `platforms` seems to be both optional and now incorrect as well: https://www.python.org/dev/peps/pep-0345/#platform-multiple-use. It can probably just be removed, since our classifiers are more granular.


setup.py
```diff
@@ -22,12 +22,13 @@
...
     tests_require=tests_require,
-    platforms='any',
     install_requires=install_requires,
     classifiers=[
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
-        'Operating System :: OS Independent',
+        'Operating System :: MacOS',
+        'Operating System :: MacOS :: MacOS X',
+        'Operating System :: POSIX :: Linux',
```
thomasst: Why not also include `Operating System :: POSIX`? I am not aware of any Linux-specific code that we're using. Are you?

setup.py
```diff
@@ -22,12 +22,13 @@
     long_description=long_description,
     test_suite='tests',
     tests_require=tests_require,
-    platforms='any',
     install_requires=install_requires,
     classifiers=[
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
-        'Operating System :: OS Independent',
+        'Operating System :: MacOS',
```
thomasst: Do we stil want this? (Since we don't support MacOS 9)

setup.py
AlecRosenbaum: Nope, I just wasn't sure what the correct tags were so I guessed.

Issue #89, opened 2017-10-12
Optimize scheduled task queueing


thomasst:
When there are many queues, `_worker_queue_scheduled_tasks` is looping through all the scheduled queues after every worker run, which causes a lot of Redis overhead.

jkemp101:
Related code:
https://github.com/closeio/tasktiger/blob/e926d7802cb3caeba0f757852e09feb5f39f16d2/tasktiger/worker.py#L1017-L1023

Some ideas:
* Provide an argument to specify the number of expected workers and do something like `self.config['SELECT_TIMEOUT']  * self.num_workers` to adjust how often it is called based on how many workers will be calling it
* Use [LimitLion](https://github.com/closeio/limitlion) to control how often workers run the related code
* Use a Redis `SETNX` call that creates a key that expires in `SELECT_TIMEOUT`. The worker that wins and sets that key makes the related calls. Need concept of worker group name so we can have a key per group of related workers (this is needed in the LimitLion approach also)

jkemp101:
For the `SETNX` or LimitLion option we could probably just have a key/throttle per root queue and not need a group name.

jkemp101:
@thomasst Should we close this? I don't think we have any further optimizations planned.

PR #162, opened 2020-04-17
Lock for scheduled tasks queueing

thomasst:
First attempt to address https://github.com/closeio/tasktiger/issues/89#issuecomment-609963574

Alternatively, we could lock for a shorter amount of time (e.g. a second) and not release the lock, which saves on Redis instructions for releasing the lock, and also prevents us from attempting to schedule tasks too often.


thomasst:
@jkemp101 Tests now pass. Did you want to play around with it on production?

PR #164, opened 2020-06-16
Fix tests timings

thomasst:
* Ąssumes processes take longer to start on  macOS Catalina

    See also https://news.ycombinator.com/item?id=23273247

    I confirmed using the commands from https://news.ycombinator.com/item?id=23281106 that it is indeed being held up by Catalina's security.

    This allows for enough time to run tests on macOS Catalina on a 2020 MacBook Pro with 2.3 GHz Quad-Core i7.

    Since many tests rely on the DELAY, and occasionally 0.2s caused issues on Circle as well, it didn't make sense to introduce a setting just for spawning the process, so I decided to increase the wait time for all tests. This doesn't seem to have a significant impact on the overall test execution time in Circle.

* Fixes timing issue in TestPeriodicTasks.test_periodic_execution. This issue happens specifically when at test start time less than 0.2 seconds are left until the next second starts, and the worker doesn't have enough time to queue periodic tasks.

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -7,7 +7,7 @@ DELAY = 0.2
 # Note that on macOS Catalina, when using an unsigned Python version,
 # taskgated (com.apple.securityd) needs to approve launching the process.
 # We therefore need ample here in order to prevent test failures.
-PROCESS_DELAY = 0.4
+PROCESS_DELAY = 2 * DELAY
 
 # Redis database number which will be wiped and used for the tests
 TEST_DB = int(os.environ.get('REDIS_DB', 7))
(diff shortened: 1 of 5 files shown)


thomasst:
I think `TestMaxWorkers.test_single_worker_queue` is still failing sometimes (https://app.circleci.com/pipelines/github/closeio/tasktiger/17/workflows/f4c4343c-b220-449f-97a8-d073e6c0f5c6/jobs/1340) but I'm not sure why.

AlecRosenbaum:
This type of issue makes me so sad 😢

PR #167, opened 2020-10-23
Add python 3.8 support and remove 2.7 and 3.5

jkemp101:
* Added tests for 3.8
* Removed Python 2.7 and 3.5 tests. Latest FreezeFrog doesn't support them
* Fixed running out of file handles when running tests locally. Wanted to switch to using fixtures but didn't have time


tsx (approved):
Any blockers with this? Can we ship it?

jkemp101:
@thomasst Any concerns removing Python 2.7 and 3.5 support. I don't think we need it anymore and we don't have any commits since v0.11 so that version would always be available for older Python users if needed.

tests/tasks.py
```diff
@@ -54,13 +57,14 @@ def long_task_ok():
...
     conn.lpush(LONG_TASK_SIGNAL_KEY, '1')
-
+    conn.close()
     time.sleep(DELAY)
 
 
 def wait_for_long_task():
     """Waits for a long task to start."""
     conn = redis.Redis(host=REDIS_HOST, db=TEST_DB, decode_responses=True)
+    conn.close()
```
thomasst, line 67: Why do you close before `blpop`?

tests/tasks.py
```diff
@@ -41,7 +42,9 @@ def task_on_other_queue():
 
 
 def file_args_task(filename, *args, **kwargs):
-    open(filename, 'w').write(json.dumps({'args': args, 'kwargs': kwargs}))
+    f = open(filename, 'w')
```
thomasst, line 45: How about we use `with open() as f`?

tests/tasks.py
```diff
@@ -54,13 +57,14 @@ def long_task_ok():
     # Signal task has started
     conn = redis.Redis(host=REDIS_HOST, db=TEST_DB, decode_responses=True)
     conn.lpush(LONG_TASK_SIGNAL_KEY, '1')
-
+    conn.close()
```
thomasst, line 60: Didn't test, but can all of these be replaced with `with Redis() as conn`?

https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L916-L920

It makes it less likely to forget closing the connection, and it also closes explicitly in case an exception happens.

thomasst (commented):
I'm OK with removing Python 2.7/3.5 support.

jkemp101:
Switched to use context managers where possible

tests/tasks.py
```diff
@@ -14,6 +14,7 @@
 LONG_TASK_SIGNAL_KEY = 'long_task_ok'
 
 tiger = get_tiger()
+tiger.connection.close()
```
thomasst, line 17: It still looks weird to me having the connection closed right after initializing TaskTiger.

What if `TaskTiger`'s `init()` would not use the Redis connection until it's needed? I believe the reason it's using it is for the scripts in https://github.com/closeio/tasktiger/blob/master/tasktiger/tasktiger.py#L224.

We could make `scripts` lazy:

```
class TaskTiger:   # side note: inheriting from object is no longer needed in Python 3
    def init(...):
        ...
        self._scripts = None

    @property
    def scripts(self):
        if self._scripts is None:
            self._scripts = RedisScripts(self.connection)
        return self._scripts
```

jkemp101: I removed that close. I don't think it was a big cause of leaked connections and I can run tests locally without it. Would rather not make more changes in this PR not relevant to changing Python version support. I just started doing it because I couldn't run the tests locally to test 3.8 and refused to use `ulimit` to make tests pass.

setup.py
```diff
@@ -31,12 +31,10 @@
         'Operating System :: POSIX :: Linux',
         'Topic :: Software Development :: Libraries :: Python Modules',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
```
thomasst, line 37: Remove the section in this file that says `if sys.version_info < (3, 3):`.

thomasst, line 37: Investigate comment on top of this file that says `# workaround for open() with encoding='' python2/3 compability`.

thomasst (approved):
Feel free to merge this as is.

I found two more things that could be addressed.

setup.py
jkemp101: Removed

jkemp101: Removed the `import os`. Looks like Py2 built-in `open` didn't support `encoding` parameter but now `io.open()` is just an alias for the built-in function.

PR #169, opened 2020-12-14
Include original task id with batched parameters

jkemp101:
This allows clients to know which task ID a set of parameters came from. Useful for logging purposes.

### DIFF
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -298,7 +298,9 @@ class Task(object):
 
         try:
             if is_batch_func:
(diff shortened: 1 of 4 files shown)


thomasst:
Would `tiger.current_tasks[N].id` be sufficient to do this? For a single task you'd also use `tiger.current_task.id` to get the ID.

thomasst:
What do we want to decide here?

jkemp101:
I started an alternative implementation using your suggestion. I removed you as a reviewer on this one.

PR #170, opened 2020-12-16
Store tiger instance in global task context

thomasst:
This makes it easier for libraries (e.g. FlowFish) to access the TaskTiger instance to queue further tasks.

```
from tasktiger._internal import g

def my_task(...):
    tiger = g["tiger"]
    tiger.delay(more_tasks)
```

I don't like that it's importing `_internal` but wasn't sure if there's a good way to expose it, perhaps just `from tasktiger import g`?

### DIFF
--- a/tasktiger/_internal.py
+++ b/tasktiger/_internal.py
@@ -174,5 +174,6 @@ class classproperty(property):
 
     Works like @property but on classes.
     """
+
     def __get__(desc, self, cls):
         return desc.fget(cls)
--- a/tasktiger/_internal.py
+++ b/tasktiger/_internal.py
@@ -166,3 +166,13 @@ def queue_matches(queue, only_queues=None, exclude_queues=None):
         if part in only_queues:
             return True
     return not only_queues
+
+
+class classproperty(property):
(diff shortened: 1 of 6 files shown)


jkemp101 (commented):
I think we should keep `g` hidden so we can refactor without breaking things. Expose it with a helper function or class method on TaskTiger? And do we need to ensure the Redis connections are closed before the child process uses it since we would be reusing the attached Redis instance from the parent process? I thought we did that somewhere but could find it with a quick search.

thomasst:
The Redis connection is closed before forking in `_execute`, so we should be good. I added a class property, `TaskTiger.current_instance`.

jkemp101 (commented):
Add a test or two and do we need to set it here also for eager tasks?

https://github.com/closeio/tasktiger/blob/f6ed7b50a34e460a558d6253b79a3c4a5e7c5d33/tasktiger/task.py#L292-L297

thomasst:
Good catch with eager tasks.

jkemp101 (approved):

PR #171, opened 2020-12-22
Batch task log iterator

jkemp101:
Alternative to https://github.com/closeio/tasktiger/pull/169


jkemp101:
@thomasst here is my latest idea. What do you think about this approach?

thomasst:
Interesting approach. Would we want `tiger.current_task` to also point to the batch task?

PR #175, opened 2021-01-12
Support for custom task runners

thomasst:


tasktiger/runner.py
```diff
@@ -0,0 +1,73 @@
...
+        log.error(
+            "cannot mix multiple runner classes",
+            runner_class_paths=", ".join(str(p) for p in runner_class_paths),
+        )
+        raise ValueError("Found multiple runner classes in batch task.")
+
+    runner_class_path = runner_class_paths.pop()
+    if runner_class_path:
+        try:
+            runner_class = import_attribute(runner_class_path)
```
tsx, line 69: Return it directly from here, otherwise it'll return None a few lines below.

tsx (approved):

tasktiger/runner.py
thomasst: Thanks. I'm also re-raising the exception instead of falling back to `DefaultRunner` since executing a task without the intended runner might have weird consequences.

tests/test_base.py
```diff
@@ -1236,3 +1238,29 @@ def fake_error(msg):
...
+
+class TestRunnerClass(BaseTestCase):
+    def test_custom_runner_class_single_task(self):
+        task = self.tiger.delay(simple_task, runner_class=MyRunnerClass)
+        Worker(self.tiger).run(once=True)
+        assert self.conn.get('task_id') == task.id
+        self.conn.delete('task_id')
+        self._ensure_queues()
+
+    def test_customer_runner_class_batch_task(self):
```
jkemp101, line 1251: Probably should have a test that confirms what happens if two batch tasks have two different runners.

README.rst
```diff
@@ -337,6 +337,27 @@ The following options are supported by both ``delay`` and the task decorator:
...
   60 seconds between executions, pass ``retry_method=fixed(60, 3)``.
 
+- ``runner_class``
+
+  If given, a Python class can be specified to influence task running behavior.
+  The runner class should inherit `tasktiger.runner.BaseRunner` and implement
+  the task execution behavior. The default implementation is available in
+  `tasktiger.runner.DefaultRunner`. The following behavior can be achieved:
+
+  - Execute specific code before or after the task is executed (while forked),
```
jkemp101, line 347: NIT: Should we use the wording "in the forked child process" somewhere in this section to be very clear these runners only operate in the child? I had to think for a moment what _(while forked)_ here meant.

jkemp101 (commented):
What about eager tasks that get executed [here](https://github.com/closeio/tasktiger/blob/6064c95525f1c196c313be909d3dbb4648ef10ca/tasktiger/task.py#L292)? Should we just not allow custom runners for eager run tasks?

README.rst
thomasst: Like this? https://github.com/closeio/tasktiger/pull/175/commits/6df51f69455308abcebe19c77a350936ede46ab5

jkemp101: Yep

tests/test_base.py
thomasst: in https://github.com/closeio/tasktiger/pull/175/commits/5653f299123cc1637583946ae08cc7c1259b9000

thomasst:
> What about eager tasks that get executed [here](https://github.com/closeio/tasktiger/blob/6064c95525f1c196c313be909d3dbb4648ef10ca/tasktiger/task.py#L292)? Should we just not allow custom runners for eager run tasks?

How about we add `run_eager_task(self, task)` to the runner so it can be customized? The function would also have to differentiate between batch or not, using `getattr(func, '_task_batch', False)`. The `BaseRunner` would just throw `NotImplementedError`.

jkemp101:
>How about we add run_eager_task(self, task) to the runner so it can be customized? The function would also have to differentiate between batch or not, using getattr(func, '_task_batch', False). The BaseRunner would just throw NotImplementedError.

That sounds good. Hopefully isn't too much work to throw that in now.

thomasst:
> That sounds good. Hopefully isn't too much work to throw that in now.

No worries, done in https://github.com/closeio/tasktiger/commit/9734e6218c41a0266b330546c93cdb1693cce875

jkemp101 (approved):

PR #176, opened 2021-02-16
Poll for new tasks

thomasst:
Ability to configure a polling interval for new task to reduce load when there are many workers & tasks. Instead of subscribing to the activity channel, which results in N_WORKERS messages for each task (and then each of them polling for the task), we just poll periodically. Downside is that with polling enabled and few workers it might take up to the polling interval to pick up a new task.

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -209,9 +209,23 @@ class Worker(object):
                 self.connection.publish(self._key('activity'), queue)
                 self._did_work = True
 
-    def _wait_for_new_tasks(self, timeout=0, batch_timeout=0):
+    def _poll_for_queues(self):
         """
-        Check activity channel and wait as necessary.
+        Refresh list of queues.
+
+        Wait if we did not do any work.
+
+        This is only used when using polling to get queues with queued tasks.
+        """
+        if not self._did_work:
+            time.sleep(self.config["POLL_QUEUED_TASKS"])
+        self._refresh_queue_set()
+
+    def _pubsub_for_queues(self, timeout=0, batch_timeout=0):
+        """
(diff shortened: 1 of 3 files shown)


tasktiger/worker.py
```diff
@@ -222,6 +222,14 @@ def _wait_for_new_tasks(self, timeout=0, batch_timeout=0):
            3. Timeout seconds have passed, this is the maximum time to stay in
               this method
         """
+        if not self._pubsub:
+            if self._did_work:
+                self._refresh_queue_set()
+            else:
+                if not self._queue_set:
```
jkemp101, line 229: Why are we `not self._queue_set`? Is the theory that if we already have some queues we know about we should drain the set before getting new ones? If so we should have a comment here to clarify the logic.

jkemp101 (commented):
Should we be publishing here if polling is enabled?
https://github.com/closeio/tasktiger/blob/6064c95525f1c196c313be909d3dbb4648ef10ca/tasktiger/worker.py#L209

tasktiger/worker.py
thomasst: I think I can remove the check since the condition `not self._did_work and self._queue_set` should not happen.

In `_worker_run` we call `_process_from_queue` for each queue, which takes a task (or multiple with batch queues) and processes it. It removes the queue from `_queue_set` if it's empty.

In order for `_did_work` to be false, we must have not processed any tasks (or scheduled any tasks). In this case we would have looped through all the queues in the `_queue_set` and removed them.

thomasst:
@jkemp101 Take another look. Logic should be equivalent.

jkemp101 (approved):

PR #177, opened 2021-02-16
Reduce worker idle queries

thomasst:
This reduces idle ops with N workers from from `6 + 4 * (N-1)` to  `7 + 1 * (N-1)` per second.


tasktiger/worker.py
```diff
@@ -160,6 +160,17 @@ def _filter_queues(self, queues):
             )
         ]
 
+    def _worker_perform_secondary_tasks(self):
+        # We should queue scheduled tasks every QUEUE_SCHEDULED_TASKS_TIME time
+        # and expired tasks every REQUEUE_EXPIRED_TASKS_INTERVAL time. Use only
+        # one Redis query to enter this block since every single worker calls
+        # this every second.
+        # XXX: Ideally, we should keep track of workers and take turns.
+        key = self._key("lock", "secondary_tasks")
```
jkemp101, line 169: Switch to single quotes

jkemp101 (commented):
Code looks fine. Would probably want to do a test with a bunch of scheduled queues to see what performance improvement it makes to make sure we understand how this helps.

PR #179, opened 2021-02-24
Add latest python and redis test

jkemp101:
This tested Python 3.9.2 and Redis 6.2.0 automatically.

Should we keep this test to check if changes are forward compatible?

/cc @AlecRosenbaum

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,6 +7,7 @@ workflows:
       - test-3.6
       - test-3.7
       - test-3.8
+      - test-latest
       - black
 
 defaults: &defaults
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,6 +7,7 @@ workflows:
       - test-3.6
       - test-3.7
       - test-3.8
+      - test-latest
```
thomasst, line 10: Shouldn't we explicitly list the supported version, e.g. `test-3.9`?

jkemp101: I think we should review the python 3.9 change log, add a `test-3.9` specific test, and update docs to officially support 3.9. But the purpose of `test-latest` could be to potentially find code changes that aren't forward compatible and require further investigation. Or could just be annoying because it fails due to something you don't have time to deal with.

PR #181, opened 2021-03-21
Include queue name in task ID generation

thomasst:
This avoids a "task not found" scenario where we queue a unique task into several queues.

For example, if we queue a unique task with ID X into queues A and B, and the task X finishes on queue A first before being processed by queue B, then the task ID X would still be queued in queue B, but the worker would not be able to find the task since the task key is deleted.

We check (https://github.com/closeio/tasktiger/blob/f48fe03b7c845d032e9f2bdc01d1939f2a6db8ff/tasktiger/task.py#L262-L278) whether the task is in a different state (so you can e.g. requeue/schedule a unique task into the same queue without losing it), but it would not be reasonable to check all other queues.

I believe this shouldn't be a completely incompatible change to roll out since uniqueness isn't the same as locking as you can still queue a unique task even if one is being executed. Its main purpose is to reduce load. TaskTiger doesn't make any guarantees that unique tasks are not executed concurrently, however, in its current implementation it doesn't do that and instead schedules the unique task for later execution if one is already running (https://github.com/closeio/tasktiger/blob/f48fe03b7c845d032e9f2bdc01d1939f2a6db8ff/tasktiger/worker.py#L797-L804).

### DIFF
--- a/tasktiger/_internal.py
+++ b/tasktiger/_internal.py
@@ -57,14 +57,19 @@ def gen_id():
     return binascii.b2a_hex(os.urandom(32)).decode('utf8')
 
 
-def gen_unique_id(serialized_name, args, kwargs):
+def gen_unique_id(queue, serialized_name, args, kwargs):
     """
     Generates and returns a hex-encoded 256-bit ID for the given task name and
     args. Used to generate IDs for unique tasks or for task locks.
     """
     return hashlib.sha256(
         json.dumps(
-            {'func': serialized_name, 'args': args, 'kwargs': kwargs},
+            {
+                'queue': queue,
+                'func': serialized_name,
+                'args': args,
+                'kwargs': kwargs,
+            },
             sort_keys=True,
         ).encode('utf8')
     ).hexdigest()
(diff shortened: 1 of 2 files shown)


tasktiger/_internal.py
```diff
@@ -57,16 +57,20 @@ def gen_id():
...
     Generates and returns a hex-encoded 256-bit ID for the given task name and
     args. Used to generate IDs for unique tasks or for task locks.
     """
+    data = {
+        'func': serialized_name,
+        'args': args,
+        'kwargs': kwargs,
+    }
+    if queue is not None:
+        data['queue'] = queue
```
AlecRosenbaum, line 71: Any reason not to serialize an explicit null?

AlecRosenbaum (approved):

tasktiger/_internal.py
thomasst: We currently use the same function for to compute the lock ID (in which case we pass `None` for the queue). It would break existing locks if we serialized an explicit null. I can add a comment for this.

jkemp101 (approved):
Looks like the `unique` docs might already suggest this behavior. Should we clarify that the lock is across all queues https://github.com/closeio/tasktiger#task-options.

thomasst:
Holding off with this PR for now since it can cause issues with existing deployments: If users rely on canceling scheduled tasks by ID, this will break.

PR #182, opened 2021-03-22
Drop some unnecessary requirements

jpmelos:
We don't support Python 2.7 anymore, so we don't need `contextlib2` or `six` anymore. Also, added a limitation on what versions it can install of `redis-py`: `redis-py>=2,<4`, since version 4 can be, in theory, breaking.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,3 @@
 click==7.1.2
-contextlib2==0.5.5; python_version < '3.3'
 redis==3.5.3
-six==1.15.0
 structlog==20.1.0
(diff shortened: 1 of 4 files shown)


thomasst (approved):
Can you also remove the contextlib check here? https://github.com/closeio/tasktiger/blob/master/tasktiger/worker.py#L28-L31

jpmelos:
Ah good catch. Done in https://github.com/closeio/tasktiger/pull/182/commits/f0a7eb67cf75c2e323e052701ba1d28e439e86be.

PR #188, opened 2021-03-24
Clear retried executions from successful periodic tasks

thomasst:
Previously, if you had a periodic task that retries up to N times for valid reasons, we would never clear the executions, so N would accumulate over the lifetime of a periodic task (forever). This PR doesn't completely fix the issue of tracking executions for previously failed periods, but at least it prevents executions from accumulating if the periodic task completes successfully after retries.

### DIFF
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -274,16 +274,23 @@ class Task(object):
 
         if not to_state:  # Remove the task if necessary
             if self.unique:
-                # Only delete if it's not in any other queue
-                check_states = {ACTIVE, QUEUED, ERROR, SCHEDULED}
-                check_states.remove(from_state)
                 # TODO: Do the following two in one call.
+
+                # Delete executions if there were no errors.
+                if from_state == ERROR:
+                    check_states = {}
+                else:
+                    check_states = {ERROR}
                 scripts.delete_if_not_in_zsets(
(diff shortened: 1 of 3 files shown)


AlecRosenbaum:
Do tasks only exit the ERROR state on retries?

Would this also apply to unique retryable tasks that aren't periodic? (I don't see an issue with this, but just clarifying)

thomasst:
> Do tasks only exit the ERROR state on retries?

Tasks can be in multiple states. Once a task is in ERROR state it doesn't go out of it. That's why I mentioned "This PR doesn't completely fix the issue" above.


> Would this also apply to unique retryable tasks that aren't periodic? (I don't see an issue with this, but just clarifying)

This wasn't an issue before. They would get cleaned up as expected as long as there wasn't another scheduled occurrence of that task.

AlecRosenbaum:
So for periodic tasks at what point do they get into `_move` where `from_state == ERROR`?

thomasst:
> So for periodic tasks at what point do they get into `_move` where `from_state == ERROR`?

If deleted from the admin or via `task.delete()`.

AlecRosenbaum (approved):

jpmelos (approved):

jkemp101 (approved):

PR #189, opened 2021-03-26
Redis SCAN script with TTL setting capabilities

jpmelos:


scripts/redis_scan.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # This is a string because we want the `while` below to run, and the string
+    # will be correctly interpreted by `client.scan` as a request for a new
+    # cursor.
+    cursor = '0'
+
+    signal.signal(signal.SIGINT, signal_handler)
+
+    with open(file_name, 'w') as log_file:
+        while cursor != 0 and loop:
```
tsx, line 80: Use `client.scan_iter()` instead. No need to mess with cursors manually.

scripts/redis_scan.py
```diff
@@ -0,0 +1,108 @@
...
+        print(f'Scanning redis keys with match: {match}\n')
+    else:
+        print('Scanning all redis keys\n')
+
+    # This is a string because we want the `while` below to run, and the string
+    # will be correctly interpreted by `client.scan` as a request for a new
+    # cursor.
+    cursor = '0'
+
+    signal.signal(signal.SIGINT, signal_handler)
```
tsx, line 77: Maybe copy-past this?

https://github.com/closeio/closeio/blob/b6e52b01b753b56fdb4ececb6c0c7cea06387419/closeio/app/utils.py#L966

scripts/redis_scan.py
```diff
@@ -0,0 +1,108 @@
...
+
+            for key in data:
+                key_type = client.type(key)
+                size = get_size(client, key, key_type)
+
+                ttl = client.ttl(key)
+                new_ttl = None
+                if set_ttl == -1:
+                    if ttl != -1:
+                        client.persist(key)
```
tsx, line 91: Can you make this, `ttl` and the call in `get_size` into a pipeline? 1 IO per key is better/faster than 3.

scripts/redis_scan.py
jpmelos: I'd rather not go that way. We want to control pressure on Redis by not issuing too many requests too fast, so we avoid starving other requests, and the in-between of `SCAN` iterations seems like a natural place to do that. We also want to flush `log_file` to disk often so that, in case something happens, we will have more accurate data on what happened.

Of course that can be achieved by other means, like counting number of keys processed and adding a delay and flushing the file every 1,000 keys, for example. But then, if I'm going to add that logic, might as well just keep the cursor logic we already have. Another reason to not do that is that, if the `match` used to look for a key has a very sparse occurrence, we could end up firing a lot of requests in rapid succession to Redis before we process 1,000 keys, so it doesn't guarantee that we won't starve other requests.

Makes sense?

scripts/redis_scan.py
jpmelos: I think that wouldn't work. We don't get size for all types of keys, only some. I'd need to add a check to see whether we ran any call to get the size. Then, the execution of `persist` depends on the result of `ttl`. Even if I removed that restriction and made the pipeline execute the `persist` command if we just want to remove TTL, in the second branch I'd need to execute the pipeline with a single command and get the TTL value to decide. Overall, I don't see much of a gain in using pipelines in this script, unless I'm missing something.

PR #191, opened 2021-04-12
Add Task.time_last_queued

froxCZ:
Needed for https://github.com/closeio/tasktiger-admin/pull/22

btw. where could I find some notes about release process of this lib?

### DIFF
--- a/tasktiger/task.py
+++ b/tasktiger/task.py
@@ -144,9 +144,11 @@ class Task(object):
 
     @property
     def time_last_queued(self):
-        return datetime.datetime.utcfromtimestamp(
-            self._data["time_last_queued"]
-        )
+        timestamp = self._data.get("time_last_queued")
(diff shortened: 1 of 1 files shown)


tasktiger/task.py
```diff
@@ -142,6 +142,12 @@ def id(self):
     def data(self):
         return self._data
 
+    @property
+    def time_last_queued(self):
+        return datetime.datetime.utcfromtimestamp(
```
thomasst, line 147: Let's return `None` if `self._data` doesn't contain it (which it could if it's not a queued task).

froxCZ: Thx, done

thomasst (approved):

thomasst:
> btw. where could I find some notes about release process of this lib?

See https://github.com/closeio/tasktiger/pull/192

Issue #178, opened 2021-02-17
Add endpoint that identifies queues with tasks but not in config


jkemp101:
When using the [ONLY_QUEUES](https://github.com/closeio/tasktiger/blob/6d96607bbe49f4c4791772cc9880bcb77a78068e/tasktiger/tasktiger.py#L187) configuration option, it is likely expected that all queues will be specified in that list. A missing queue could mean tasks left unprocessed since be default workers would ignore that queue. This is a silent failure until it is noticed by an  operator or other queue alerts trigger.

We should add a `would_process_queue(queue_name)` method to the [`TaskTiger`](https://github.com/closeio/tasktiger/blob/f48fe03b7c845d032e9f2bdc01d1939f2a6db8ff/tasktiger/tasktiger.py#L76) class  that accepts a queue name and uses the instance's config and `queue_matches()` to determine if the specified queue would be processed by this tiger instance. It should handle subqueues so that if `ONLY_QUEUES` has `xyz` and you check queue `xyz.123` it would return `True`.

Monitoring agents can then verify the queues returned by `get_queues_stats` with `would_process_queue()` function to determine if there is a configuration error.

PR #194, opened 2021-04-16
Add TaskTiger.would_process_configured_queue

froxCZ:
Closes https://github.com/closeio/tasktiger/issues/178

### DIFF
--- a/tasktiger/tasktiger.py
+++ b/tasktiger/tasktiger.py
@@ -591,6 +591,12 @@ class TaskTiger(object):
 
         return total_processed
 
+    def would_process_queue(self, queue_name):
+        """
+        Return if a queue_name would be processed by this tasktiger instance. It does not consider explicit
+        --queue param.
+        """
+        return queue_matches(queue_name, self.config['ONLY_QUEUES'], self.config['EXCLUDE_QUEUES'])
 
 @click.command()
 @click.option(
--- a/tasktiger/tasktiger.py
+++ b/tasktiger/tasktiger.py
@@ -591,12 +591,17 @@ class TaskTiger(object):
 
         return total_processed
 
-    def would_process_queue(self, queue_name):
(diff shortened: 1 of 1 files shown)


tasktiger/tasktiger.py
```diff
@@ -591,6 +591,12 @@ def errored_tasks():
 
         return total_processed
 
+    def would_process_queue(self, queue_name):
+        """
+        Return if a queue_name would be processed by this tasktiger instance. It does not consider explicit
+        --queue param.
```
froxCZ, line 597: @thomasst Ignoring the --queue param can be confusing and don't like it much. On the other hand, we would have to store the worker reference, and then use `worker.only_queues if worker else config['ONLY_QUEUES']` Thoughts?

thomasst: I would not store the worker reference on the `TaskTiger` object. 

It seems like the specific `--queue` parameters are currently specific to the worker. We could:

- Make the TaskTiger instance responsible for managing queues. For example, we could only initialize when we have to command line args (which might be tricky), or, have a function like  `set_queues_for_processing()` which is called to set/update the queues we want to process. The worker would then not store a set of queues but rely on the TaskTiger reference.
- Leave it as you did in the PR: The TaskTiger instance knows which queues it can process in general (from the config), but the Worker instance may have specific inclusions/exclusions.

The first option would be my preferred one but is also more work to do. @jkemp101 Do you have a preference here?

Also note max line length should be 79.

froxCZ: Agreed with Joe to just rename to `would_process_configured_queue`

Ideal state would be workers reporting on which queues they are subscribed, and the monitoring agent would use that info instead of checking the config. But that would also be more work, so doing the simple thing now.

thomasst (approved):

PR #201, opened 2021-12-02
Add queue name to logs

drewler:
Log queue name of the current task from the default structlog processor.


tasktiger/logging.py
```diff
@@ -9,6 +9,8 @@ def tasktiger_processor(logger, method_name, event_dict):
     """
```
thomasst, line 9: Docstring:
```
Inject the current task ID and queue for non-batch tasks.
```

thomasst (approved):

PR #205, opened 2022-03-16
Bump python versions

AlecRosenbaum:
drop 3.6 where it's mentioned (in the docker image + circleci testing), add 3.9 to circleci


thomasst (approved):
Also feel free to add 3.10

AlecRosenbaum:
I initially tried, but some of the dependencies are out of date and don't yet work on 3.10. I think we should upgrade dependencies separately then revisit adding 3.10 testing

PR #216, opened 2022-03-30
Update circleci to use the new images

AlecRosenbaum:
Use `cimg/` instead of `circleci/`. This also bumps the redis version to 5.x, which is the minimum version supported by circleci's new image set.

CI failures are unrelated and are already fixed in https://github.com/closeio/tasktiger/pull/215


thomasst (approved):

thomasst:
The tests are failing though.

PR #242, opened 2022-11-30
Prefilter polled queues

neob91-close:
This is a PoC for reducing the Redis load caused by each worker frequently polling the entire set of subqueues.

SSCAN lets us prefilter the queues we pull out of Redis, and seems pretty fast too:
```py
production In [49]: r.scard("t:scheduled")
production Out[49]: 3919

production In [50]: timeit.timeit(lambda: r.smembers("t:scheduled"), number=5)
production Out[50]: 0.06490249998751096

production In [51]: timeit.timeit(lambda: r.sscan("t:scheduled", match="zoom.*", count=10000), number=5)
production Out[51]: 0.01752800800022669
```

### DIFF
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -24,6 +24,7 @@ from .task import Task
 from .timeouts import JobTimeoutException
 
 LOCK_REDIS_KEY = "qslock"
+REDIS_GLOB_CHARACTER_PATTERN = re.compile(r"([?*\[\]])")
 
 __all__ = ["Worker"]
 
@@ -1165,19 +1166,15 @@ class Worker(object):
         if len(self.only_queues) != 1:
             return self.connection.smembers(key)
 
-        # Escape special characters in the queue
-        match = re.sub(r"([?*\[\]])", r"\\\1", list(self.only_queues)[0]) + "*"
-
-        result = set()
-        cursor = None
-
-        while cursor != 0:
-            cursor, items = self.connection.sscan(
-                key, cursor=cursor or 0, match=match, count=100000
+        # Escape special Redis glob characters in the queue name
+        match = (
+            REDIS_GLOB_CHARACTER_PATTERN.sub(
+                r"\\\1", list(self.only_queues)[0]
             )
-            result.update(items)
+            + "*"
+        )
 
-        return result
+        return set(self.connection.sscan_iter(key, match=match, count=100000))
 
     def run(self, once=False, force_once=False):
         """
--- a/tasktiger/worker.py
+++ b/tasktiger/worker.py
@@ -3,6 +3,7 @@ from contextlib import ExitStack
 import errno
 import fcntl
 import random
+import re
 import select
 import signal
 import socket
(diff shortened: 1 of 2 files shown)


tasktiger/worker.py
```diff
@@ -1156,9 +1157,27 @@ def _queue_periodic_tasks(self):
...
+        self._queue_set = set(self._filter_queues(self._retrieve_queues()))
+
+    def _retrieve_queues(self):
+        key = self._key(QUEUED)
+
+        if len(self.only_queues) != 1:
+            return self.connection.smembers(key)
+
+        # Escape special characters in the queue
+        match = re.sub(r"([?*\[\]])", r"\\\1", list(self.only_queues)[0]) + "*"
```
vtclose, line 1169: Compile the regex ahead of time

tasktiger/worker.py
```diff
@@ -1156,9 +1157,27 @@ def _queue_periodic_tasks(self):
...
+
+        result = set()
+        cursor = None
+
+        while cursor != 0:
+            cursor, items = self.connection.sscan(
```
vtclose, line 1175: `sscan_iter`

vtclose (commented):
A couple nitpicks but I'm not very knowledgeable with TT internals to comment on correctness.

tasktiger/worker.py
neob91-close: Neat! 😅

vtclose (approved):

neob91-close:
I'm going to deploy this on canary next week in the morning (EU time).

tasktiger/worker.py
```diff
@@ -23,6 +24,7 @@
 from .timeouts import JobTimeoutException
 
 LOCK_REDIS_KEY = "qslock"
+REDIS_GLOB_CHARACTER_PATTERN = re.compile(r"([?*\[\]])")
```
thomasst, line 27: The relevant characters to be escaped are `*`, `?`, `[`, and `\` (https://github.com/redis/redis/blob/155acef51ac7826ed294a4b61d891f1c7a9a40ac/src/util.c#L58). You additionally have `]` (which is fine but irrelevant), but you should also escape `\`, so this should be `re.compile(r"([?*\[\]\\])")` (or `re.compile(r"([?*\[\\])")`).

thomasst (approved):
Looks good except the comment.

tasktiger/worker.py
neob91-close: I don't know how I overlooked the `\` character, thanks!

I am aware that it is not necessary to escape `]`, but to my knowledge Redis docs do not mention that fact and I prefer to avoid depending on undocumented behaviour.

thomasst: Which docs btw? I was trying to find Redis docs on globs but couldn't find any, so then I just looked at the source code.

thomasst (approved):

tasktiger/worker.py
vtclose: This has some examples https://redis.io/commands/keys/ but doesn't seem like a formal spec

drewler (approved):

PR #243, opened 2022-12-01
Sort imports

neob91-close:
I noticed there's no linting other than `black` in CI. We might want to gradually start adding it so we could eventually start using `lintlizard`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,15 @@ exclude = '''
     \.git
 )/
 '''
+
+[tool.isort]
+skip = ['.git', 'venv']
+known_tests = ['tests']
+sections = ['FUTURE', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'TESTS', 'LOCALFOLDER']
+default_section = 'THIRDPARTY'
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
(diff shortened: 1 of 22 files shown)


jpmelos (approved):

pyproject.toml
```diff
@@ -5,3 +5,15 @@ exclude = '''
...
+skip = ['.git', 'venv']
+known_tests = ['tests']
+sections = ['FUTURE', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'TESTS', 'LOCALFOLDER']
+default_section = 'THIRDPARTY'
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
+multi_line_output = 3
+line_length = 78
+float_to_top = true
```
jpmelos, line 19: This is a bit different from what is in `closeio`. Is that intentional? Line length is 79 in `closeio`, and we also have `combine_as_imports = true`.

jpmelos: Sorry for not flagging this as part of my initial review, I only had the idea to compare them after I had already submitted. 😬

neob91-close: That's a good point. I actually scavenged and revised the config from a repository that has a slightly different config than our ones. I'll open another PR to bring this one in sync.

Issue #249, opened 2022-12-05
Add Type Checking


AlecRosenbaum:
Type checking is a good idea for a number of reasons. It helps with a number of bug classes, and also serves to document the code.

We should configure `mypy` for this project and enable type checking within CI. We should then gradually introduce typing to this project.

PR #251, opened 2022-12-07
Allow truncating task executions

neob91-close:
**IMPORTANT**
This is a breaking change warranting a major version bump and a breaking changes section in the changelog!

---

This PR introduces a way to configure the limit of executions stored at `t:task:<uuid>:executions`.
We still need to know the total number of times the task has run (for retries), so we're storing the total number of executions in  `t:task:<uuid>:executions_count`.

Redis RPUSH used to add executions to the list is replaced with a script called `TRUNCATED_RPUSH` which removes the left elements as needed.


nsaje (commented):
LGTM! let's hear what @thomasst thinks

neob91-close:
I just realised that we could keep the execution counter inside the execution payload itself instead of maintaining a second counter key. I'm putting this on hold until I hear some feedback about this particular idea and the PR in general.

tasktiger/task.py
```diff
@@ -286,30 +298,38 @@ def _move(self, from_state=None, to_state=None, when=None, mode=None):
...
-                    check_states = {}
-                else:
-                    check_states = {ERROR}
-                scripts.delete_if_not_in_zsets(
-                    _key("task", self.id, "executions"),
-                    self.id,
-                    [_key(state, queue) for state in check_states],
-                    client=pipeline,
-                )
+                if from_state != ERROR:
```
neob91-close, line 301: Just realised that we should probably just delete these keys if `from_state == ERROR`.

thomasst (commented):
> I just realised that we could keep the execution counter inside the execution payload itself instead of maintaining a second counter key. I'm putting this on hold until I hear some feedback about this particular idea and the PR in general.

Is `executions_truncation_counter` just needed to report the total number of executions? If so, I think it would make more sense to just call it `executions_count`, since it's more useful to know the total number of executions rather than amount of executions that got truncated.

Should the `max_tracked_executions` be a per-task setting or a global setting in the config? I almost feel like this is something that would be specific to a deployment rather than the task itself. Or are we thinking we just want to apply it to specific tasks?

tasktiger/redis_scripts.py
```diff
@@ -473,13 +490,15 @@ def srem_if_not_exists(self, key, member, other_key, client=None):
             keys=[key, other_key], args=[member], client=client
         )
 
-    def delete_if_not_in_zsets(self, key, member, set_list, client=None):
+    def delete_if_not_in_zsets(self, to_delete, value, zsets, client=None):
```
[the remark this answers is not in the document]
neob91-close: https://github.com/closeio/tasktiger/issues/249
Got it on my todo-list

neob91-close:
> > I just realised that we could keep the execution counter inside the execution payload itself instead of maintaining a second counter key. I'm putting this on hold until I hear some feedback about this particular idea and the PR in general.
> 
> Is `executions_truncation_counter` just needed to report the total number of executions? If so, I think it would make more sense to just call it `executions_count`, since it's more useful to know the total number of executions rather than amount of executions that got truncated.

Yes. Did it this way, because it felt more normalised this way, but storing the complete count is fine with me too.

> Should the `max_tracked_executions` be a per-task setting or a global setting in the config? I almost feel like this is something that would be specific to a deployment rather than the task itself. Or are we thinking we just want to apply it to specific tasks?

For some tasks (webhook delivery) we should expect a significant number of retries. For other tasks we don't really need/want to truncate executions.

tasktiger/redis_scripts.py
```diff
@@ -506,6 +522,22 @@ def get_expired_tasks(self, key_prefix, time, batch_size, client=None):
         # [queue1, task1, queue2, task2] -> [(queue1, task1), (queue2, task2)]
         return list(zip(result[::2], result[1::2]))
 
+    def truncated_rpush(self, key, element, max_length, client=None):
```
thomasst, line 525: Just use a pipeline with `rpush` and `ltrim -max_length -1`.

neob91-close: bfd5b7b6da67beec141dd277f539f96f50b2ce09

tests/test_base.py
```diff
@@ -269,6 +269,28 @@ def test_exception_task(self, store_tracebacks):
...
+        assert _get_stored_executions() == 0
+
+        for __ in range(6):
+            Worker(self.tiger).run(once=True)
+            time.sleep(DELAY)
+
+            Worker(self.tiger).run(once=True)
+
+        assert task.n_executions() == 6
+        assert _get_stored_executions() == max_stored_executions or 6
```
thomasst, line 292: This assert is true, regardless of the values of `_get_stored_executions()` and `max_stored_executions`  (and also not what you intended even if it was parenthesized as `(max_stored_executions or 6)`).

I believe you want this:
```
assert _get_stored_executions() == min(max_stored_executions or 6, 6)
```

thomasst (approved):
Looks all good except my comment.

tests/test_base.py
neob91-close: I think it originally made sense, but the lack of parentheses prevented me from noticing when it ceased to make sense.

Fixed, thanks.

thomasst (approved):

PR #252, opened 2022-12-12
Add task execution counter

neob91-close:
Introduces a task execution counter that's meant to be used instead of counting the number of executions using Redis `LLEN`.

This is the first step towards being able to truncate the list of executions.

### DIFF
--- a/tasktiger/migrations.py
+++ b/tasktiger/migrations.py
@@ -1,3 +1,6 @@
+from .utils import redis_glob_escape
+
+
 def migrate_executions_count(tiger):
     """
     Backfills ``t:task:<uuid>:executions_count`` by counting
@@ -13,8 +16,9 @@ def migrate_executions_count(tiger):
         """
     )
 
-    for key in tiger.connection.scan_iter(
-        count=100, match="*:task:*:executions"
-    ):
-        if key.startswith(tiger.config["REDIS_PREFIX"] + ":task:"):
-            migrate_task(keys=[key, key + "_count"])
+    match = (
+        redis_glob_escape(tiger.config["REDIS_PREFIX"]) + ":task:*:executions"
+    )
+
+    for key in tiger.connection.scan_iter(count=100, match=match):
+        migrate_task(keys=[key, key + "_count"])
--- a/tasktiger/migrations.py
+++ b/tasktiger/migrations.py
@@ -1,0 +1,20 @@
+def migrate_executions_count(tiger):
+    """
+    Backfills ``t:task:<uuid>:executions_count`` by counting
+    elements in ``t:task:<uuid>:executions``.
+    """
+
+    migrate_task = tiger.connection.register_script(
+        """
+        local count = redis.call('llen', KEYS[1])
+        if tonumber(redis.call('get', KEYS[2]) or 0) < count then
+            redis.call('set', KEYS[2], count)
+        end
+        """
+    )
+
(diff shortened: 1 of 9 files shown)


tasktiger/migrations.py
```diff
@@ -0,0 +1,18 @@
+def migrate_executions_count(tiger):
+    """
+    Backfills ``t:task:<uuid>:executions_count`` by counting
```
neob91-close, line 3: We'll mention this in the changelog. Should be sufficient, right?

AlecRosenbaum: this probably warrants a `0.17` release + a big bold "**Breaking Change**" headline

vtclose (commented):
Would you mind shipping the `black` update separately?

tasktiger/migrations.py
neob91-close: Yes, I was planning a major version bump. Also, I agree about making sure that the users are aware so that nothing works in an unexpected way for them.

That being said, this PR doesn't actually introduce any breaking changes, it just adds the counter.
The actual breaking changes will be introduced in a subsequent PR.

Also, if someone misses this info and they upgrade directly to the change from the next PR, the worst that would happen is that the retry counter would reset for all tasks started on a worker before this change.

AlecRosenbaum: > the worst that would happen is that the retry counter would reset for all tasks started on a worker running on a version from before this change

Just pointing out that depending on the use case this could be very problematic. It's not outlandish to have a task where the execution/retry logic needs to be in sync with some sort of retryable business logic that cares about how many times it should attempt something before giving up/falling back to permanent failure logic.

neob91-close: I'm not saying it couldn't be problematic and I'm definitely not saying we should ignore this. It's obviously a breaking change.

I was just pointing out (for no practical reason, I guess) that there's a good chance that the users who don't read the changelog before upgrading will not even encounter any issues.

tasktiger/migrations.py
```diff
@@ -0,0 +1,20 @@
...
+        """
+        local count = redis.call('llen', KEYS[1])
+        if tonumber(redis.call('get', KEYS[2]) or 0) < count then
+            redis.call('set', KEYS[2], count)
+        end
+        """
+    )
+
+    for key in tiger.connection.scan_iter(
+        count=100, match="*:task:*:executions"
```
thomasst, line 17: Should the `match` just include the `REDIS_PREFIX` alreaady and then we don't have to do the check below?

thomasst (approved):
- Can you document the new key in the large docstring on top of `tasktiger/tasktiger.py`?
- I'm assuming we'll change `Task.n_executions()` in a later PR for compatibility.

tasktiger/migrations.py
neob91-close: 8405c95314c94cf6f95e4ed2eecd25d402e0b01a

neob91-close:
> Can you document the new key in the large docstring on top of tasktiger/tasktiger.py?

Done in 8405c95314c94cf6f95e4ed2eecd25d402e0b01a.

> I'm assuming we'll change Task.n_executions() in a later PR for compatibility.

Yes, this is changed in the second PR: https://github.com/closeio/tasktiger/pull/251
I was planning to make a minor release after merging this PR and then a major release after merging the second PR.

PR #254, opened 2022-12-14
Introduce lintlizard and typechecking

neob91-close:
Lays the foundation for https://github.com/closeio/tasktiger/issues/249


tasktiger/runner.py
```diff
@@ -85,6 +85,9 @@ def get_runner_class(log, tasks):
         try:
             return import_attribute(runner_class_path)
         except TaskImportError:
-            log.error("could not import runner class", func=retry_func)
```
neob91-close, line 88: `retry_func` is not defined

tasktiger/worker.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         single_worker_queues=None,
         max_workers_per_queue=None,
         store_tracebacks=None,
-    ):
+    ) -> None:
```
AlecRosenbaum, line 55: mypy needs return types on all of these but doesn't care about arg types?

AlecRosenbaum (approved):
LGTM, though in the future this type of change would be easier to review if it was split up between a few commits. E.g.
* add lintlizard + setup.cfg
* fix object inheritance
* add return types
* ...

tasktiger/worker.py
neob91-close: I only added the return types so that `mypy` would do at least basic checks of some functions.
We'll need to add more typing later and then gradually make the config stricter.

PR #258, opened 2022-12-16
Fix some flake8 issues

neob91-close:
Improves imports and fixes some miscellaneous flake8 issues.

Could theoretically break things for users, but only if someone is importing tasktiger internals via `from tasktiger import ...`, however I'm not sure whether we should actually treat this as a breaking change, because these things weren't meant to be exposed in the main module.


AlecRosenbaum (approved):
I don't think this is a breaking change since `__init__.py` already has an `__all__` which hasn't really changed.

thomasst (approved):

PR #265, opened 2023-03-01
Add 'task_func' to logging processor

neob91-close:
This would be useful for searching all log entries related to a given task function.

I also took the liberty of doing a slight refactor of `tests.test_logging`.

### DIFF
--- a/tasktiger/logging.py
+++ b/tasktiger/logging.py
@@ -11,6 +11,7 @@ def tasktiger_processor(logger, method_name, event_dict):
     if g["current_tasks"] is not None and not g["current_task_is_batch"]:
         current_task = g["current_tasks"][0]
         event_dict["task_id"] = current_task.id
+        event_dict["task_func"] = current_task.serialized_func
         event_dict["queue"] = current_task.queue
 
     return event_dict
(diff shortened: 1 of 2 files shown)


tasktiger/logging.py
```diff
@@ -11,6 +11,7 @@ def tasktiger_processor(logger, method_name, event_dict):
     if g["current_tasks"] is not None and not g["current_task_is_batch"]:
         current_task = g["current_tasks"][0]
         event_dict["task_id"] = current_task.id
+        event_dict["task_func"] = current_task.serialized_func
```
vtclose, line 14: Why not reuse the same `"func"` key as in the logs done by TT itself?

vtclose (approved):

tasktiger/logging.py
neob91-close: I'm not comfortable introducing a global `func` label because it may clash with app logs.
It's not difficult to imagine that someone somewhere uses `logger.info("foo", func="...")`.

neob91-close: Ideally I'd have all of these prefixed with `tasktiger`, but I don't want to mess with that too much ;-)

PR #271, opened 2023-03-27
StatsThread updates

neob91-close:
Changes:
- Replace `time.time()` with `time.monotonic()` for time measurements.
- Get rid of `threading.Event.isSet` call, because `threading.Event.wait` already returns the same exact value.

### DIFF
--- a/tasktiger/stats.py
+++ b/tasktiger/stats.py
@@ -11,7 +11,7 @@ class StatsThread(threading.Thread):
         self._stop_event = threading.Event()
 
         self._task_running = False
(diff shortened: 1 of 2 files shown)


AlecRosenbaum (approved):
Did anything in particular prompt the move to `monotonic`?

neob91-close:
Just a thing I noticed while working on something else. `time.monotonic()` is preferred to `time.time()` when it comes to time measurements, because it's immune to changes and calibrations of the system clock.

PR #279, opened 2023-04-19
Expose stats

neob91-close:
This allows hooking into the stats by defining a `STATS_CALLBACK` which will receive task runtimes.
This is useful for measuring the utilisation of workers.


tasktiger/stats.py
```diff
@@ -64,9 +78,34 @@ def compute_stats(self) -> None:
...
+    def _record_time_busy(self, now: float) -> None:
+        assert self._task_start_time
+        self._time_busy += now - max(self._task_start_time, self._time_start)
+
+
+class StatsLoggingThread(threading.Thread):
+    def __init__(self, stats: Stats) -> None:
+        super().__init__()
+
+        self.tiger = stats.tiger
```
thomasst, line 100: Should this ideally just say `self._stats_interval = stats.tiger.config["STATS_INTERVAL"]` rather than looking it up every time?

tasktiger/tasktiger.py
```diff
@@ -198,6 +198,12 @@ def init(
             "BATCH_QUEUES": {},
             # How often to print stats.
             "STATS_INTERVAL": 60,
+            # The function to call with the duration of each completed task.
+            # This can be useful to measure the worker's utilisation %.
+            # For example, the worker's utilisation over the last 30 minutes
+            # can be obtained by dividing the sum of task durations reported
+            # over the last 30 minutes by 30 minutes.
+            "STATS_CALLBACK": None,
```
thomasst, line 206: Since the point of the stats thread is to print out metrics periodically, I would have also expected this to be called periodically rather than at the end of the task, and report the same metrics that we log (time total, time busy, utilization). Otherwise this could be solved via something like `CHILD_CONTEXT_MANAGERS`, although that runs in the child process. Should this be `TASK_END_STATS_CALLBACK` then?

> the worker's utilisation over the last 30 minutes can be obtained by dividing the sum of task durations reported over the last 30 minutes by 30 minutes.

This is not actually correct. Example: If a task runs from 0m to 5m, and another task runs from 29m59s-34m59s, and the second task's stats callback looks at any reports in the past 30 minutes (i.e. 5m - 34m59s) , it would show 10m in task durations over 30m (~33% utilization), rather than the actual 5m1s (~17% utilization).

thomasst (commented):
Would like to know some more details on how this will be used exactly before approving.

tasktiger/tasktiger.py
vtclose: We should definitely have both start and end callbacks to measure and report utilization accurately while the task is still running. With our configuration, we often collect metrics on sub-minute intervals, and a single task can run for more than a few minutes.

neob91-close: > Since the point of the stats thread is to print out metrics periodically, I would have also expected this to be called periodically rather than at the end of the task, and report the same metrics that we log (time total, time busy, utilization). 

Yep, that sounds better than what I did.

> This is not actually correct. Example: If a task runs from 0m to 5m, and another task runs from 29m59s-34m59s, and the second task's stats callback looks at any reports in the past 30 minutes (i.e. 5m - 34m59s) , it would show 10m in task durations over 30m (~33% utilization), rather than the actual 5m1s (~17% utilization).

Good point.

> We should definitely have both start and end callbacks to measure and report utilization accurately while the task is still running. With our configuration, we often collect metrics on sub-minute intervals, and a single task can run for more than a few minutes.

Couldn't we achieve it with simply moving the callback to the `log` method to be called at an interval?
I don't think we need to get notified of the start and stop of every task.

PR #280, opened 2023-04-20
Fix Flask-Script integration

neob91-close:
Importing `tasktiger.flask_script` blows up:

```py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/src/tasktiger/flask_script.py", line 12, in <module>
    class TaskTigerCommand(Command):
  File "/src/tasktiger/flask_script.py", line 21, in TaskTigerCommand
    def __init__(self, tiger: TaskTiger) -> None:
NameError: name 'TaskTiger' is not defined
```

### DIFF
--- a/tasktiger/flask_script.py
+++ b/tasktiger/flask_script.py
@@ -18,7 +18,7 @@ class TaskTigerCommand(Command):
     capture_all_args = True
     help = "Run a TaskTiger worker"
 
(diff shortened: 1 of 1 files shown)


thomasst (approved):
Also looks like we want to remove this entirely.

Issue #268, opened 2023-03-14
Support Redis >= 6.2.7 (Lua readonly global tables)


nsaje:
On redis >= 6.2.7 we currently encounter an "Attempt to modify a readonly table" error (more info [here](https://github.com/closeio/tasktiger/issues/238))
This error is introduced by https://github.com/redis/redis/pull/10651 which hardened Lua script execution by not allowing modifications of global tables and is affecting Redis versions >= 6.2.7.

It happens because when trying to execute a pipeline using the `execute_pipeline` Lua script, we try to set the `KEYS` and `ARGV` global variables before calling another script using `EVALSHA`.

Executing pipelines via a Lua script was [introduced](https://github.com/closeio/tasktiger/issues/54) because Redis transactions don't abort execution of commands when one of the commands in the transaction fails.

The fix would probably be to use [Redis functions](https://redis.io/docs/manual/programmability/functions-intro/) for Redis 7.0 and above, because they're directly callable by providing `keys` and `argv` in the call itself using `FCALL <func_name> <keys> <argv>`

nsaje:
cc @thomasst , any thoughts on this? 😄

thomasst:
> The fix would probably be to use [Redis functions](https://redis.io/docs/manual/programmability/functions-intro/) for Redis 7.0 and above, because they're directly callable by providing `keys` and `argv` in the call itself using `FCALL <func_name> <keys> <argv>`

I have not played around with Redis 7 but that looks good. The way we're doing it currently is a hack.

joeleclems:
Hi,
For information, we encountered the same problem with redis 6.2.6. We also have redis database in version 6.2.5, and no problems with theses bdd.
So i think the problem is on redis >= 6.2.6

PR #284, opened 2023-06-15
Rewrite `_move` helper in Lua (to add Redis 7 support)

nsaje:
closes https://github.com/closeio/tasktiger/issues/268

In order to not need the `execute_pipeline` script that doesn't work on Redis >= 6.2.7 (and also on earlier versions with security patches backported, like on AWS ElastiCache), I rewrote the `_move()` helper in Lua directly.

Added the ability to include existing scripts as local functions when loading a script from file in order to facilitate code reuse.

Any suggestions how to improve the Lua code? For example Redis doesn't know how to pass bools and nulls to Lua so I had to convert them to strings, any better options?


tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,85 @@
...
+    zadd_w_mode(key(to_state, queue), when, id, mode)
+    redis.call('sadd', key(to_state), queue)
+end
+redis.call('zrem', key(from_state, queue), id)
+
+if to_state == "" then -- Remove the task if necessary
+    if unique == 'true' then
+        -- Delete executions if there were no errors
+        local to_delete = {
+            key('task', id, 'executions'),
```
neob91-close, line 49: Since we're rewriting this from scratch, we shouldn't be generating keys inside lua scripts.

From [Redis docs](https://redis.io/docs/manual/programmability/eval-intro/):
```
At this point, it is essential to understand the distinction Redis makes between input arguments that are names of keys and those that aren't.

While key names in Redis are just strings, unlike any other string values, these represent keys in the database. The name of a key is a fundamental concept in Redis and is the basis for operating the Redis Cluster.

Important: to ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments. The script should only access keys whose names are given as input arguments. Scripts should never access keys with programmatically-generated names or based on the contents of data structures stored in the database.
```

tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,85 @@
+local function zadd_w_mode(key, score, member, mode)
+    if mode == "" then
+        redis.call('zadd', key, score, member)
+    elseif mode == "nx" then
+        zadd_noupdate({ key }, { score, member })
```
neob91-close, line 5: Couldn't we just call `zadd` with `nx`?

tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,85 @@
...
+    elseif mode == "xx" then
+        zadd_update_existing({ key }, { score, member })
```
neob91-close, line 7: Couldn't we just call `zadd` with `xx`?

tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,85 @@
...
+    elseif mode == "min" then
+        zadd_update_min({ key }, { score, member })
```
neob91-close, line 9: Sounds like we could just call `zadd` with `lt`?

tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,85 @@
...
+    elseif mode == "max" then
+        zadd_update_max({ key }, { score, member })
```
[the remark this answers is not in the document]
neob91-close: Though `lt` and `gt` are not supported by Redis 4.0, so we should think whether we can drop support here.

(`nx` and `xx` are supported though)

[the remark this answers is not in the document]
nsaje: replying just once, applies to all :smile:

two reasons I decided against it:
- NX and XX were added in v3.0.2, LT & GT were added in 6.2.0. we're currently running TT on 3.2 so we can't use LT&GT
- I wanted to keep the implementation as close as possible to the original, just rewritten to Lua. That's because this change is risky as is, I don't want to do optimizations and refactors in the same step as well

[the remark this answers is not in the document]
nsaje: an example for the 2nd point - the native implementation of NX, XX, GT _could_ be subtly different from what we do in our scripts. It could slip through and work most of the time but cause bugs or corruption down the line.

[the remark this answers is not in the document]
neob91-close: I checked, and `lt`/`gt` were introduced somewhere after Redis 6.0 (probably in Redis 7.0), so we'd need to drop support for Redis 4, 5 and 6. Probably not the best idea.

neob91-close (commented):
The change looks correct in general, except the keys should be passed as the `KEYS` input.

tasktiger/task.py
```diff
@@ -308,75 +307,26 @@ def _move(
...
-        if to_state == QUEUED and self.tiger.config["PUBLISH_QUEUED_TASKS"]:
-            pipeline.publish(_key("activity"), queue)
-
         try:
-            scripts.execute_pipeline(pipeline)
+            scripts.move_task(
+                key_prefix=self.tiger.config["REDIS_PREFIX"],
+                id=self.id,
+                queue=self.queue,
+                from_state=from_state or self.state,
```
thomasst, line 322: `from_state` is sufficient since above you say `from_state = from_state or self.state`.

tasktiger/lua/move_task.lua
nsaje: done in latest commit!

tasktiger/redis_scripts.py
```diff
@@ -330,11 +346,24 @@ def can_replicate_commands(self) -> bool:
...
+        self, filename: str, include_functions: Optional[dict] = None
+    ) -> Script:
         with open(
             os.path.join(os.path.dirname(os.path.realpath(__file__)), filename)
         ) as f:
-            return self.redis.register_script(f.read())
+            script = f.read()
+            if include_functions:
+                function_definitions = []
+                for func_name, func_body in include_functions.items():
```
thomasst, line 358: Since this is a dict, the order isn't technically guaranteed (although I believe the order is consistent in later Python versions). Just to be safe we should sort the keys to avoid storing multiple similar script with different function orders in Redis.

tasktiger/lua/move_task.lua
```diff
@@ -0,0 +1,95 @@
+local function zadd_w_mode(key, score, member, mode)
```
thomasst, line 1: "nx", "min", and "" are the only ones we actually use, right? So we could technically skip including "xx" and "max".

tasktiger/redis_scripts.py
nsaje: great point, will fix

tasktiger/lua/move_task.lua
nsaje: true, will trim it

thomasst (approved):

PR #293, opened 2023-08-09
docs: remove secret from public repo

meMuszr:
Looks like token is no longer valid, however it's been in the codebase for 8 years. Not sure when repo went public, but best to remove and fix the status badge not displaying correctly anyway.


AlecRosenbaum (approved):
Thank you for the cleanup!

PR #296, opened 2023-08-24
add tiger.current_serialized_func

nsaje:
Adds a `current_serialized_func` property to the `tiger` object.

This is useful for instrumentation - the `current_serialized_func` can be used for naming spans or labeling metrics.

Without it one would have to do e.g.

```python
try:
    return tiger.current_task.serialized_func
except RuntimeError:
    return tiger.current_tasks[0].serialized_func
```

because we don't expose `current_task_is_batch` either.


thomasst (approved):
I guess we could expose `current_task_is_batch` also if needed.

PR #309, opened 2023-11-11
Refactor Redis keys to avoid duplicate string literals

### LINKED ISSUE #255: Refactor keys to avoid duplicate string literals
We're currently using a lot of string literals for building Redis keys, and it's relatively easy to make a typo that might go unnoticed.
It would be nice if we could improve on this, maybe use a separate method for generating each key?

vishesh10:
## Description
Replaced string literals in the Redis keys to constants.

Resolves #255


tests/test_base.py
```diff
@@ -127,7 +138,7 @@ def test_simple_task(self):
 
         Worker(self.tiger).run(once=True)
         self._ensure_queues(queued={"default": 0})
-        assert not self.conn.exists("t:task:%s" % task["id"])
+        assert not self.conn.exists(f"{REDIS_PREFIX}:{TASK}:%s" % task["id"])
```
neob91-close, line 141: I would keep these literals unchanged in tests to make it clearer what these keys look like.

neob91-close (changes requested):
This seems to almost only change tests (which we probably don't want to change) and not the literals in the actual feature code.

vishesh10:
Thanks for the review. 
If tests are skipped then there are very few changes in the code.

neob91-close:
There are a lot of places that need changes. A couple of examples:
- https://github.com/closeio/tasktiger/blob/b1019e667c3a199481c518752258d723d4dcd836/tasktiger/task.py#L396
- https://github.com/closeio/tasktiger/blob/b1019e667c3a199481c518752258d723d4dcd836/tasktiger/worker.py#L377

There's probably a couple dozen lines that need an update.

vishesh10:
@neob91-close,
Reverted the test cases and added changes at the appropriate places.
Thanks

tasktiger/constants.py
```diff
@@ -0,0 +1,11 @@
+# constants pertaining to Redis keys
+REDIS_PREFIX = "t"
+TASK = "task"
+EXECUTIONS = "executions"
+EXECUTIONS_COUNT = "executions_count"
+
+
+QUEUED = "queued"
```
neob91-close, line 8: These are already defined in `tasktiger._internal`.
What's more, these definitions aren't even used.
We should get rid of them.

tasktiger/constants.py
```diff
@@ -0,0 +1,11 @@
+# constants pertaining to Redis keys
+REDIS_PREFIX = "t"
```
neob91-close, line 2: This isn't used, nor is it needed (it's configurable, and the default value is defined in `tasktiger.tasktiger`.

tasktiger/constants.py
```diff
@@ -0,0 +1,11 @@
...
+TASK = "task"
```
neob91-close, line 3: Can we move these definitions to `tasktiger._internal` for consistency?

neob91-close (changes requested):
There's some more changes necessary.
By the way, sorry it took me so long to get back to you.

vishesh10:
Hi @neob91-close ,
Done changes as per the suggestions.
Thanks

PR #320, opened 2024-02-15
Synchronous executor

thomasst:
Not sure if that's the best name for it -- basically just an executor that runs tasks without threads or processes. It should handle hard timeouts via signal from the runner. Can use `-x sync` or `-x fork` (default) to specify which executor to use.

Can review by commit.

Quick benchmark:
```
class TestSyncExecutorWorkerPerformance:
    def test_sync(self, tiger, ensure_queues):
        import time

        ensure_queues()

        for n in range(1000):
            Task(tiger, simple_task).delay()

        ensure_queues(queued={"default": 1000})

        worker = Worker(tiger, executor_class=SyncExecutor)

        start_time = time.time()
        worker.run(once=True)
        end_time = time.time()

        ensure_queues()
        print("sync took", end_time - start_time)

    def test_fork(self, tiger, ensure_queues):
        import time

        ensure_queues()

        for n in range(1000):
            Task(tiger, simple_task).delay()

        ensure_queues(queued={"default": 1000})

        worker = Worker(tiger)

        start_time = time.time()
        worker.run(once=True)
        end_time = time.time()

        ensure_queues()
        print("fork took", end_time - start_time)
```

gives:
```
tests/test_workers.py::TestSyncExecutorWorkerPerformance::test_sync sync took 0.4465901851654053
PASSED
tests/test_workers.py::TestSyncExecutorWorkerPerformance::test_fork fork took 4.435322046279907
PASSED
```

But the other consideration will be checking out connections from an existing connection pool rather than having to re-establish them.

### DIFF
--- a/tasktiger/executor.py
+++ b/tasktiger/executor.py
@@ -152,6 +152,10 @@ class Executor:
 
             self.worker.store_task_execution(tasks, execution)
 
+        g["current_task_is_batch"] = None
+        g["current_tasks"] = None
+        g["tiger"] = None
+
         return success
 
 
--- a/tasktiger/executor.py
+++ b/tasktiger/executor.py
@@ -80,11 +80,88 @@ class Executor:
             locks: List of task locks to renew periodically.
             queue_lock: Optional queue lock to renew periodically for max
                 workers per queue.
+
+        Returns:
+            Whether task execution was successful.
         """
         raise NotImplementedError
 
+    def execute_tasks(self, tasks: List[Task], log: BoundLogger) -> bool:
+        """
+        Executes the tasks in the current process. Multiple tasks can be passed
+        for batch processing. However, they must all use the same function and
+        will share the execution entry.
+        """
+        success = False
+
+        execution: Dict[str, Any] = {}
+
+        assert len(tasks)
+        task_func = tasks[0].serialized_func
+        assert all([task_func == task.serialized_func for task in tasks[1:]])
+
+        execution["time_started"] = time.time()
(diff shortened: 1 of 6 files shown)


thomasst:
Do we need to restart the process if a `UnixSignalDeathPenalty` occurs? (i.e. exit with error)

tasktiger/tasktiger.py
```diff
@@ -691,6 +694,11 @@ def would_process_configured_queue(self, queue_name: str) -> bool:
 @click.option("-p", "--port", help="Redis server port")
 @click.option("-a", "--password", help="Redis password")
 @click.option("-n", "--db", help="Redis database number")
+@click.option(
+    "-x",
```
tsx, line 698: `-x` can potentially be confused with `-e` as in "eXclude" a queue. I'd keep just the long option.

tasktiger/executor.py
```diff
@@ -80,11 +80,88 @@ def execute(
...
+        execution["time_started"] = time.time()
+
+        try:
+            func = tasks[0].func
+
+            runner_class = get_runner_class(log, tasks)
+            runner = runner_class(self.tiger)
+
+            is_batch_func = getattr(func, "_task_batch", False)
+            g["tiger"] = self.tiger
```
tsx, line 112: Would it make sense to unset that (and other parts of `g`) at the end of the execution?

tsx (approved):

tsx (approved):

PR #328, opened 2024-03-05
Fix Python language classifiers

thomasst:
Adjusting the Python version matrix to what's being tested.

Do we even need them?


AlecRosenbaum (approved):
I have no idea if the classifiers actually do anything for us

PR #370, opened 2026-03-24
Support for task workflow/dependencies

mku11:
I'm planning on using tasktiger as a light-weight CI workflow scheduler. These are the basic changes that support this concept if you're interested.

Changes:
* New optional parameter `depends` for class `Task` that will be used to assign a list of task ids as dependencies.
* 2 new states: `WAITING` and `COMPLETED`.
* When a task is moved to `ACTIVE` state and dependencies have not completed it will move to the `WAITING` state.
* When a task is completed it will move any tasks that are dependent from `WAITING` to `SCHEDULED` state.
* NOTE: All tasks after completion will remain in `COMPLETED` state instead of removing from db!
* New optional parameter `max_parallel_workers` for `TaskTiger.run_worker()` and command line option `-P --max-parallel-workers`. This is a simple wrapper that will fork multiple workers for user convenience.
* Tests cases are updated, I had to add pytz to the requirements file to make it work with github ci.

This is based on a squash merge from my fork.
I'll also be creating a PR with the tasktiger-admin changes for supporting these changes.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,4 @@
 freezefrog==0.4.1
 psutil==5.9.8
 pytest==8.1.1
+pytz
--- a/tasktiger/_internal.py
+++ b/tasktiger/_internal.py
@@ -33,6 +33,8 @@ QUEUED = "queued"
 ACTIVE = "active"
 SCHEDULED = "scheduled"
 ERROR = "error"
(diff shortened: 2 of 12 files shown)


mku11:
This is the corresponding changes for tasktiger-admin: https://github.com/closeio/tasktiger-admin/pull/31

PR #379, opened 2026-08-13
Prevent swallowed timeouts from wedging sync workers

AlecRosenbaum:
This PR prevents sync workers from remaining indefinitely wedged when asynchronous task handling swallows a hard-timeout exception. It retries timeout delivery before terminating the worker so its supervisor can replace it.


tasktiger/timeouts.py
```diff
@@ -43,7 +44,15 @@ def cancel_death_penalty(self) -> None:
 
 
 class UnixSignalDeathPenalty(BaseDeathPenalty):
+    def __init__(self, timeout: float) -> None:
+        super().__init__(timeout)
+        self.retries = 0
+
     def handle_death_penalty(self, signum: int, frame: Any) -> None:
+        if self.retries >= 3:
+            sys.exit(1)
```
AlecRosenbaum: should I use `os._exit` instead?

thomasst (approved):
I'm assuming we deliberately don't want to do any cleanup here (such as `threading._shutdown()` etc), so `_exit` is probably fine. Are we able to log or print an error safely here?

